PR #3, opened 2020-12-01
Boilerplate

tony-iqlusion:
Assorted CI and repo-level configuration:

- Define MSRV (1.48.0) and add `rust-toolchain` file
- Run clippy in CI using MSRV toolchain
- Use `RUSTFLAGS` to deny warnings in CI; allow them locally


cosmos-sdk-proto/src/lib.rs
```diff
@@ -5,8 +5,7 @@
 //! worth considering that modules seem to show up and go away with every RC.
 //! TODO actually implement features tag based compilation
 
-#![deny(warnings, trivial_casts, trivial_numeric_casts, unused_import_braces)]
```
tony-iqlusion, line 8: I'm a fan of allowing warnings locally, but enforcing code is warning-free in CI. See the corresponding change here:

https://github.com/cosmos/cosmos-rust/pull/3/files#diff-73e17259d77e5fbef83b2bdbbe4dc40a912f807472287f7f45b77e0cbf78792dR15

.github/workflows/rust.yml
```diff
@@ -2,26 +2,40 @@ name: Rust
 
 on:
   push:
-    branches: [master]
+    branches: master
+    paths-ignore:
+      - README.md
   pull_request:
-    branches: [master]
+    branches: master
```
tony-iqlusion, line 9: I think this is preventing GH Actions from running... going to remove it

zmanian (approved):

cosmos-sdk-proto/src/lib.rs
```diff
@@ -5,8 +5,7 @@
...
-#![allow(clippy::large_enum_variant)]
```
tony-iqlusion, line 9: This clippy lint used to have an annoyingly small default but was increased to ~200 bytes somewhat recently, which I think is a reasonable threshold

greg-szabo (approved):

Issue #4, opened 2020-12-02
Enable compilation of server-side gRPC files


adizere:
By default, we're only building client-side gRPC files:

https://github.com/cosmos/cosmos-rust/blob/388b18b73e513e9ac7896cf8d245c68ea4c57bf4/proto-build/src/main.rs#L212-L213

Towards providing support for both client-side and server-side implementations, shall we consider adding a cmd-line argument to enable compilation of server-side files?

This feature is needed in a mock chain implementation by @CharlyCst, comment below:

----

I was about to open an issue for a feature request, but it seems that it would be more appropriate to post it here:
I'm building a mock node for testing purpose and I rely on `ibc-proto` for proto definitions, however `proto-compiler` only builds client-related definitions (see [this line](https://github.com/informalsystems/ibc-rs/blob/2aa305bfc2079708e3d7cc77595d5587b4c67821/proto-compiler/src/cmd/compile.rs#L117)). For now I simply forked the crate but I would very much appreciate having all the definitions in a single place.

I'm not very familiar with `ibc-proto` and the needs of the cosmos ecosystem in general, but if someone is willing to provide some guidance I can help on moving thins forward.

_Originally posted by @CharlyCst in https://github.com/informalsystems/ibc-rs/issues/425#issuecomment-735920505_

jkilpatr:
I think this should be pretty straightforward. I'll give it a try later today.

adizere:
closing as stale, we no longer need this right now

PR #5, opened 2020-12-03
WIP: Enable compliation of server side definitions

jkilpatr:
This is a very simple patch to enable compilation of server definitions
for the gRPC services we're compiling. I don't think this adds enough
bloat to justify the complexity of a compile time flag.

I also think that this feature as requested in #4  is clearly in scope. 

All that being said this pull is a WIP because of errors in the generated code that need to be resolved.


adizere:
Any thoughts on this?

tony-iqlusion:
Seems like a good feature, but the PR is incomplete (not passing tests) and needs to be rebased

jkilpatr:
this is stale, closing.

PR #6, opened 2020-12-07
cosmos-tx crate

tony-iqlusion:
Adds a `cosmos-tx` crate which provides a transaction builder and signer for Cosmos-based blockchains.

This commit ports over the WIP protobuf support from:

https://github.com/iqlusioninc/crates/tree/develop/stdtx

### DIFF
--- a/.github/workflows/cosmos-sdk-proto.yml
+++ b/.github/workflows/cosmos-sdk-proto.yml
@@ -1,0 +1,25 @@
+name: cosmos-sdk-proto
+
+on:
+  push:
+    branches: main
+  pull_request:
+    paths:
+      - "cosmos-sdk-proto/**"
+      - "Cargo.*"
+
+defaults:
+  run:
+    working-directory: cosmos-sdk-proto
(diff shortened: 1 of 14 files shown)


tony-iqlusion:
For repos with multiple crates like this, I think it's helpful to have separate CI jobs for each crate. Here's an example:

https://github.com/RustCrypto/AEADs/tree/master/.github/workflows

I didn't do that yet, but I think it'd be a good idea.

jkilpatr (approved):
This looks preggy good to me as a start. I agree some testing changes need to be made, find with multiple workflows for testing each crate.

Let me know once this is ready for an integration test and I'll submit one that spins up a cosmos blockchain for testing against.

tony-iqlusion:
I can split the workflows and then it should be PoC-ready, if you'd like to handle integration testing in a separate PR.

jkilpatr:
yup sounds good, integration testing should be separate as it's going to be big enough to review on it's own.

PR #7, opened 2020-12-07
Add Dependabot config

tony-iqlusion:
Automatically opens PRs to update out-of-date dependencies:

https://docs.github.com/en/free-pro-team@latest/github/managing-security-vulnerabilities/configuring-dependabot-security-updates

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+version: 2
+updates:
+- package-ecosystem: cargo
+  directory: "/"
+  schedule:
+    interval: weekly
+  open-pull-requests-limit: 10


.github/dependabot.yml
```diff
@@ -0,0 +1,7 @@
+version: 2
+updates:
+- package-ecosystem: cargo
+  directory: "/"
+  schedule:
+    interval: weekly
```
tony-iqlusion, line 6: `weekly` helps reduce overall churn, especially for dependencies that release frequently

jkilpatr:
In my experience dependabot opens a lot of random PR's especially in projects with lots of dependencies. 

Why do we need dependabot versus just not specifying the patch version in our dependencies list? With the exclusion of patch versions most things don't get updated but every several months.

tony-iqlusion:
Since Cargo.lock is checked in, Dependabot also ensures we have the latest versions in Cargo.lock.

We can turn down the settings to reduce the overall number of PRs, for example reducing the number of concurrently open PRs.

.github/dependabot.yml
```diff
@@ -0,0 +1,7 @@
...
+  open-pull-requests-limit: 10
```
tony-iqlusion, line 7: This can be adjusted down if desired.

jkilpatr:
> Since Cargo.lock is checked in, Dependabot also ensures we have the latest versions in Cargo.lock.

Wouldn't this technically require opening a pull request any time any transitive dependency changed? I don't think this is feasible just from a sheer scale perspective, you have update a lot of transitive deps at once otherwise we'll be looking at hundreds of PRs.

.github/dependabot.yml
jkilpatr: I think 5 is good. Keep things on major versions only and don't automatically update the cargo.lock, there's no substitute for knowing what deps tree everyone is using in production. If you don't keep track you'll at some point regret it.

tony-iqlusion:
I believe Dependabot will only upgrade immediate dependencies as specified in `Cargo.toml`, but does the equivalent of `cargo update -p <dependency>`, so it will upgrade transitive dependencies indirectly as part of the PRs it opens.

jkilpatr (approved):

PR #8, opened 2020-12-08
cosmos-sdk-proto: `grpc` feature + Test WASM builds in CI

tony-iqlusion:
Adds `grpc` feature to `cosmos-sdk-proto` crate allowing it to be built without gRPC support, enabling usage in WASM environments.

Also adds CI jobs which ensures that both `cosmos-sdk-proto` and `cosmos-tx` are WASM-friendly.


tony-iqlusion:
Looks like we'd need some Cargo features to disable Tonic/gRPC for this to work.

proto-build/src/main.rs
```diff
@@ -256,3 +267,15 @@ fn copy_generated_files(from_dir: &Path, to_dir: &Path) {
...
+    let contents = fs::read_to_string(src)?;
+
+    // Patch each service definition with a feature attribute
+    let patched_contents = contents.replace(
+        TONIC_SERVICE_ATTRIBUTE,
+        &format!("{}\n{}", GRPC_FEATURE_ATTRIBUTE, TONIC_SERVICE_ATTRIBUTE),
+    );
+
+    fs::write(dest, patched_contents)
+}
```
tony-iqlusion, line 281: Hax, but unfortunately while I saw `tonic_build` lets you add attributes to generated structs/enums, it doesn't appear to provide any way to easily add an attribute to modules.

cosmos-sdk-proto/Cargo.toml
```diff
@@ -10,16 +10,16 @@ readme     = "README.md"
...
-
 [dependencies]
 prost = "0.6"
 prost-types = "0.6"
-tonic = "0.3.1"
-git2 = "0.13"
-prost-build = "0.6"
-tempdir = "0.3"
-walkdir = "2"
-tonic-build = "0.3.1"
```
tony-iqlusion, line 22: These appear to be superfluous (moved to `proto-build`?)

jkilpatr: yes they are for proto-build, good catch.

proto-build/src/main.rs
jkilpatr: pretty clever I like it.

jkilpatr (approved):
Generally happy, would kinda like separate commits for different subjects. But that's whatever.

PR #9, opened 2020-12-08
proto-build: set `publish = false` in Cargo.toml

tony-iqlusion:
Prevents accidental publication to crates.io


jkilpatr:
I don't think anyone was going to try this, but probably best.

jkilpatr (approved):

PR #10, opened 2020-12-08
Fix GitHub Actions badges

tony-iqlusion:
The "Rust" workflow was split into separate workflows-per-crate in #6.

However, the badges in the README.md files were not updated. This commit updates them to point to the new workflows.


tony-iqlusion:
This seems minor enough I'm just going to go ahead and merge

PR #12, opened 2020-12-14
proto-build: use git submodules for managing cosmos-sdk repo

tony-iqlusion:
Normally I hate git submodules, but this seems like a valid use case.

This commit removes the use of the `git2` crate for checking out the cosmos-sdk repo and uses a git submodule instead.

The previous usages of `git2` are replaced with a small automation for invoking the `git submodule update --init` command, which passes through the git status printed to standard output.

It also avoids recloining the repository every time `proto-build` is run.


cosmos-sdk-proto/src/prost/COSMOS_SDK_COMMIT
```diff
@@ -1 +1 @@
-v0.40.0-rc3
\ No newline at end of file
+590358652cc1cbc13872ea1659187e073ea38e75
```
tony-iqlusion, line 1: The main drawback of this approach, at least as implemented right now, is it's sourcing this value from `.git/modules/cosmos-sdk/HEAD` which contains a raw commit hash instead of a nicely formatted tag name.

See the updated `output_sdk_version` below:

https://github.com/cosmos/cosmos-rust/pull/12/files#diff-58406f56aa9701745b57c8af4adb067401ed9cb97ee7cad24938a905f8314f36R63-R66

jkilpatr (changes requested):
I think the removal of COSMOS_REV is not acceptable. If the goal is to make generating proto code easier and more predictable then this needs several modifications to bring it closer to the behaviour of the git2 method.

What if you edit the files in your submodule, then you run proto-gen. Those edits won't be upstream anywhere for anyone to work with and you might not even know they had affected proto generation, unless you specifically cd into the folder and run git status you won't remember or know that your submodule is out of date, or that you made some changes there months ago.

In terms of reproducibility this change as-is is a big step backwards.

tony-iqlusion:
Re: reproducibility, perhaps I should mention that presently (even without these changes) I get slightly different output running `proto-build` than what's checked in (mostly whitespace). In general it'd probably be good to run `proto-build` in CI to ensure the checked-in output matches what's generated.

As for `COSMOS_REV` and preventing hand-edits to the contents of `cosmos-sdk/`, one option would be to run `git reset --hard <COSMOS_REV>` after initializing the submodule to ensure it's in a clean state and matches the revision in the code.

jkilpatr:
> Re: reproducibility, perhaps I should mention that presently (even without these changes) I get slightly different output running `proto-build` than what's checked in (mostly whitespace). In general it'd probably be good to run `proto-build` in CI to ensure the checked-in output matches what's generated.

This happens a lot for Go proto tools, but my hope is that Cargo.lock would prevent it with our Rust builder. Seems we're not so lucky, wonder how hard the difference will be to track down to get reproducible builds. 

> As for `COSMOS_REV` and preventing hand-edits to the contents of `cosmos-sdk/`, one option would be to run `git reset --hard <COSMOS_REV>` after initializing the submodule to ensure it's in a clean state and matches the revision in the code.

This would be an acceptable solution.

tony-iqlusion:
Put back `COSMOS_REV` and added an invocation of `git reset` in f66c845

jkilpatr (approved):

PR #16, opened 2020-12-17
proto-build discrepancies

tony-iqlusion:
Note: not for merge.

I've checked in the changes I get when I run `proto-build` locally to show the discrepancies I get when I run it locally.


tony-iqlusion:
Fixed in #17

PR #17, opened 2020-12-17
proto-build: format prost/tonic output

### LINKED ISSUE #15: Reproducibility problems with `proto-build`
I've pushed up a branch containing the files changed when I run `proto-build` locally:

https://github.com/cosmos/cosmos-rust/pull/16/files

It appears to be mostly whitespace changes. I'm not sure what the cause would be.

This change to `cosmos.base.abci.v1beta1.rs` in particular (edit: actually a few files) looks pretty strange:

https://github.com/cosmos/cosmos-rust/pull/16/files#diff-7380925ee3d5ab57486392d57f665066eaafd2b93b062a8dcdac1b6fa301839cR98

```
-    pub header: ::std::option::Option<tendermint_proto::types::Header>,
+    pub header: ::std::option::Option<super::super::super::tendermint::types::Header>,
```

For whatever reason it's outputting `super::super::super::tendermint` instead of `tendermint_proto`.

tony-iqlusion:
Closes #15.

Enables the `format` option of `tonic-build` in hopes of addressing the reproducibility problems noted in #15.


tony-iqlusion:
Note: 55c02b5 contains the changes to `proto-build` in isolation.

proto-build/src/main.rs
```diff
@@ -204,10 +206,22 @@ fn copy_generated_files(from_dir: &Path, to_dir: &Path) {
...
+    // defined in the `tendermint-proto` crate. Patch them to use the real crate.
+    // See: https://github.com/cosmos/cosmos-rust/issues/15
+    // TODO(tarcieri): mega hax, find a real solution
+    let contents = contents.replace(
+        MALFORMED_TENDERMINT_PROTO_CRATE_PATH,
+        "tendermint_proto"
+    ).replace(
+        "super::tendermint_proto",
+            "tendermint_proto"
+    );
```
tony-iqlusion, line 219: For whatever reason when I run `proto-build` locally the output is attempting to use relative paths to access types defined in the `tendermint-proto` crate, as I noted in #15.

I would definitely like to find a better solution than this before merging because as the comment notes it's mega hax. However, this was enough to at least make the output reproducible on my end.

proto-build/src/main.rs
```diff
@@ -27,8 +27,10 @@ const TMP_BUILD_DIR: &str = "/tmp/tmp-protobuf/";
 
 // Patch strings used by `copy_and_patch`
 
-/// Attribute preceeding a Tonic service definition
-const TONIC_SERVICE_ATTRIBUTE: &str = "# [doc = r\" Generated client implementations.\"] pub mod";
```
tony-iqlusion, line 31: Oops, client not server.

proto-build/src/main.rs
```diff
@@ -204,10 +206,18 @@ fn copy_generated_files(from_dir: &Path, to_dir: &Path) {
 fn copy_and_patch(src: impl AsRef<Path>, dest: impl AsRef<Path>) -> io::Result<()> {
     let contents = fs::read_to_string(src)?;
 
+    // For undetermined reasons `prost-build` is outputting this path for types
+    // defined in the `tendermint-proto` crate. Patch them to use the real crate.
+    // See: https://github.com/cosmos/cosmos-rust/issues/15
+    // TODO(tarcieri): mega hax, find a real solution
+    let contents = contents
+        .replace(MALFORMED_TENDERMINT_PROTO_CRATE_PATH, "tendermint_proto")
+        .replace("super::tendermint_proto", "tendermint_proto");
```
tony-iqlusion, line 215: For whatever reason when I run proto-build locally the output is attempting to use relative paths to access types defined in the tendermint-proto crate, as I noted in #15.

I would definitely like to find a better solution than this before merging because as the comment notes it's mega hax. However, this was enough to at least make the output reproducible on my end.

jkilpatr: I'm not happy with it either, but it seems to be the best solution around right now. Lets go for it.

jkilpatr (changes requested):
Just undo your last commit and split this up into part one for your code changes and part two for the generated code, much easier to review. Different pull requests too if you prefer.

Otherwise I strongly approve.

jkilpatr (approved):

jkilpatr:
I dumb, misread the force push lines didn't see the second commit. You're a go.

PR #19, opened 2020-12-17
Rename `cosmos-sdk` to `cosmos-sdk-go`

tony-iqlusion:
After looking at the `cosmos-sdk/` submodule in my editor, I realized it's not exactly clear that this directory is the Go library rather than a `cosmos-sdk` crate (which would aspirationally be a cool thing to have)

This renames it so it's clear the directory contains the upstream Go SDK.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "cosmos-sdk"]
-	path = cosmos-sdk
+	path = cosmos-sdk-go
 	url = https://github.com/cosmos/cosmos-sdk.git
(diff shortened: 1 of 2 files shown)


tony-iqlusion:
Sidebar, but I took a look at the `cosmos-sdk` crate on crates.io:

http://crates.io/crates/cosmos_sdk

It looks like @adrianbrink owns it. Hey Adrian, would you be interested in contributing the crate to this project? We're trying to build a collaborative Cosmos SDK in Rust (no plans for it yet, but it seems like something we could eventually use)

PR #24, opened 2020-12-19
Run `proto-build` automatically with GitHub Actions

tony-iqlusion:
Adds a `rebuild-protos` GitHub Actions workflow to automatically run `proto-build` every time a commit is merged/pushed to the `main` branch.

After running `rebuild-protos`, if there are any changed files, it uses the [peter-evans/create-pull-request](https://github.com/marketplace/actions/create-pull-request) GitHub Action to open a PR with those changes automatically, thus ensuring `cosmos-sdk-proto` is reproducible and automating the work of regenerating the protos so it can be kept out-of-band with other PRs.

If no files are changed, the workflow otherwise completes successfully.

Note that this is using a 3rd party GitHub Action, and exposing the repo's `GITHUB_TOKEN` to it. However, the action is pinned to the latest release (`peter-evans/create-pull-request@v3.5.2`) effectively making it trust-on-first-use (i.e. if future updates to the action introduce malware, we will not be affected unless we choose to bump versions. If however the malware is already there, we're vulnerable).

If we're really worried, we can do a quick audit of the code and/or fork the action, either of which might be wise.

### DIFF
--- a/.github/workflows/rebuild-protos.yml
+++ b/.github/workflows/rebuild-protos.yml
@@ -1,0 +1,31 @@
+name: Rebuild Protos
+
+on:
+  push:
+    branches: main
+
+jobs:
+  proto-build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions-rs/toolchain@v1
+        with:
+          toolchain: 1.48.0 # MSRV
+          components: rustfmt
+          override: true
+          profile: minimal
+      - name: Run proto-build
+        id: build
+        working-directory: proto-build
+        run: |
+          message=$(cargo run --quiet -- --github)
+          echo "::set-output name=commit_message::${message}"
+
+      - name: Create pull request
+        uses: peter-evans/create-pull-request@v3.5.2
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          commit-message: ${{ steps.assign.outputs.commit_message }}
+          title: ${{ steps.assign.outputs.commit_message }}
+          branch: rebuild-protos
--- a/proto-build/src/main.rs
+++ b/proto-build/src/main.rs
@@ -5,13 +5,19 @@
 
 use regex::Regex;
 use std::{
+    env,
     ffi::OsStr,
     fs::{self, create_dir_all, remove_dir_all},
     io,
     path::{Path, PathBuf},
     process,
+    sync::atomic::{self, AtomicBool},
 };
 use walkdir::WalkDir;
+
+/// Suppress log messages
+// TODO(tarcieri): use a logger for this
+static QUIET: AtomicBool = AtomicBool::new(false);
 
 /// The Cosmos commit or tag to be cloned and used to build the proto files
 const COSMOS_REV: &str = "v0.40.0-rc3";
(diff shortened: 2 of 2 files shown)


tony-iqlusion:
This is unfortunately the sort of PR that you can only test by merging, especially as the workflow is configured to be "on push" to `main`.

I'll just say that this is adapted from the GitHub Action we use on @RustSec to automatically assign advisory IDs after PRs are merged and it's been working quite well in that capacity.

marbar3778:
> Note that this is using a 3rd party GitHub Action, and exposing the repo's `GITHUB_TOKEN` to it. However, the action is pinned to the latest release (`peter-evans/create-pull-request@v3.5.2`) effectively making it trust-on-first-use (i.e. if future updates to the action introduce malware, we will not be affected unless we choose to bump versions. If however the malware is already there, we're vulnerable).

Don't think this is a concern as the token only lasts an hour, and if we prevent anyone from pushing to master nothing adversarial can happen.

.github/workflows/rebuild-protos.yml
```diff
@@ -0,0 +1,30 @@
+name: Rebuild Protos
+
+on:
```
marbar3778, line 3: you could test this by add pull_request: here and changing some proto file. Then we just close the opened PR

marbar3778 (approved):
utACK

Should be fine to merge but would be better to test here?

tony-iqlusion:
Re: pinning third party actions, I've discovered versions are insufficient since they're just tags that can potentially be changed in the upstream repo. The only way to pin immutably is to a commit hash.

If we're not too worried about supply chain attacks and think the branch protections are sufficient, that's fine by me as well.

.github/workflows/rebuild-protos.yml
tony-iqlusion: In theory it shouldn't make a PR if the protos are up-to-date, but I can try adding that temporarily to at least run the workflow

tony-iqlusion: I believe this error is due to the checked out branch not having the same commit hash as `HEAD`:

https://github.com/cosmos/cosmos-rust/pull/24/checks?check_run_id=2231232423

```
Checking the base repository state
  /usr/bin/git symbolic-ref HEAD --short
  fatal: ref HEAD is not a symbolic ref
  /usr/bin/git rev-parse HEAD
  2e955eca0b443bc9480d277281bf893d1ffa3420
  Working base is commit '2e955eca0b443bc9480d277281bf893d1ffa3420'
  Error: When the repository is checked out on a commit instead of a branch, the 'base' input must be supplied.
```

So I think the only way to actually test it is to merge it unfortunately

tony-iqlusion:
Going to try merging it. If it's broken and/or annoying we can revert.

tony-iqlusion:
It failed due to an empty commit message:

https://github.com/cosmos/cosmos-rust/pull/24/checks?check_run_id=2231232423

```
  Uncommitted changes found. Adding a commit.
  /usr/bin/git add -A
  /usr/bin/git -c author.name=tony-iqlusion -c author.email=tony-iqlusion@users.noreply.github.com -c committer.name=GitHub -c committer.email=noreply@github.com commit -m 
  Aborting commit due to empty commit message.
  Error: The process '/usr/bin/git' failed with exit code 1
```

Curiously it says there were uncommitted changes:

```
  Working base is branch 'main'
  /usr/bin/git checkout --progress -B 70372755-9672-4099-9494-0e724995f08e HEAD
  Switched to a new branch '70372755-9672-4099-9494-0e724995f08e'
  M	cosmos-sdk-go
  M	cosmos-sdk-proto/src/prost/ibc.applications.transfer.v1.rs
  M	cosmos-sdk-proto/src/prost/ibc.core.client.v1.rs
  M	cosmos-sdk-proto/src/prost/ibc.core.connection.v1.rs
  /usr/bin/git status --porcelain -unormal
   M cosmos-sdk-go
   M cosmos-sdk-proto/src/prost/ibc.applications.transfer.v1.rs
   M cosmos-sdk-proto/src/prost/ibc.core.client.v1.rs
   M cosmos-sdk-proto/src/prost/ibc.core.connection.v1.rs
  Uncommitted changes found. Adding a commit.
```

PR #32, opened 2021-01-04
Bump cosmos-sdk rev to v0.40.0-rc6

tony-iqlusion:
This is the commit hash from this upstream PR which fixes the proto definitions so they compile again:

https://github.com/cosmos/cosmos-sdk/pull/8236


jkilpatr (approved):
I notice the 'tendermint' files are not currently being used in the repo should we bother to check in the rust files in that case? We are depending on tendermint-proto for those definitions.

Otherwise good and the extra generated files are not offensive, just not needed.

cosmos-sdk-proto/src/prost/tendermint.p2p.rs
```diff
@@ -0,0 +1 @@
+
```
tony-iqlusion, line 1: @jkilpatr when you say:

> I notice the 'tendermint' files are not currently being used in the repo should we bother to check in the rust files in that case?

Were you talking about this file, or something else?

Clearly this file is bogus and should probably be excluded.

jkilpatr: it seems that all the files starting with tendermint.* are totally unused and not compiled.

tony-iqlusion:
Looks like cosmos-sdk v0.40.0-rc6 is imminent: https://github.com/cosmos/cosmos-sdk/pull/8249

tony-iqlusion:
`v0.40.0-rc6` is tagged! 🎉

Went ahead and bumped this PR. There were no changes to the generated protos.

PR #58, opened 2021-03-29
Remove Cargo.lock

jkilpatr:
This removes the Cargo.lock file for this repo. 

Why does this repo have a cargo.lock anyways? Won't it be ignored by consuming clients since we're pushing non-binary libraries? @tony-iqlusion


tony-iqlusion (approved):
There's a few reasons to check in `Cargo.lock`, the main one being to cache artifacts, but that's an optimization that's best measured and we aren't using any sort of caching yet. My guess is at the moment it probably wouldn't help.

The other is to have a deterministic "known working" environment to help diagnose build failures, but that's more useful when you have a lot more outside contributors who need help diagnosing those problems, and we're not there yet.

PR #59, opened 2021-03-30
rebuild-protos: fix name of build step

tony-iqlusion:
Fixes a bit of copypasta from the action I based this on.

A local test of running `cargo run --quiet -- --github` in the `proto-build` directory showed it was working:

```
$ echo $message
Rebuild protos with proto-build (cosmos-sdk rev: v0.42.3)
```

...but the step name being passed to `create-pull-request` was incorrect.


tony-iqlusion:
Since this is just fixing what's effectively a typo, I'm going to go ahead and merge it

PR #63, opened 2021-04-01
cosmos-tx: flatten API

tony-iqlusion:
The current implementation consists of a number of single-type modules which are re-exported at the toplevel.

This commit removes the `pub` from modules to eliminate the module hierarchy, which makes the relevant types easier to find at-a-glance when reading the corresponding `rustdoc` documentation.

## Before

Screen Shot 2021-04-01 at 10 14 51 AM

## After

Screen Shot 2021-04-01 at 10 16 06 AM


jkilpatr (approved):
👍🏻  LGTM

PR #64, opened 2021-04-02
Expand cosmos-sdk-proto to include all upstream proto files

jkilpatr:
This pull request expands cosmos-sdk-proto to include all files from the upstream cosmos-sdk proto folder. It also adds services relating to transactions to the new modules and existing modules where they where previously missing. This should allow consumers of the library to import use cosmos-sdk-proto for producing many different transaction types without having to trace them down themselves.

During this change I've alphabetized all the proto imports and lists wherever they exist. Purely for ease of keeping track of what modules are or are not imported.

This patch rebuilds the proto files purely for the sake of making sure the tests pass, since the proto will be auto-regenerated once it's merged. Which is a bit of a waste of a commit  but not terrible in exchange for the other benefits. I suppose we could have the tests rebuild proto before running, which would resolve this minor issue. 

I've updated the minor version of cosmos-sdk-proto, since these changes are fully backwards compatible and we've expanded the API surface without changing any existing interfaces.

### DIFF
--- a/cosmos-tx/Cargo.toml
+++ b/cosmos-tx/Cargo.toml
@@ -11,7 +11,7 @@ categories = ["cryptography", "cryptography::cryptocurrencies", "encoding"]
 keywords   = ["blockchain", "cosmos", "tendermint", "transaction"]
 
 [dependencies]
-cosmos-sdk-proto = { version = "0.3", default-features = false, path = "../cosmos-sdk-proto" }
+cosmos-sdk-proto = {version = "0.4", default-features = false, path = "../cosmos-sdk-proto" }
 ecdsa = { version = "0.10", features = ["std"] }
 eyre = "0.6"
 k256 = { version = "0.7", features = ["ecdsa", "sha256"] }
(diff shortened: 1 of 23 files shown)


cosmos-tx/Cargo.toml
```diff
@@ -11,7 +11,7 @@ categories = ["cryptography", "cryptography::cryptocurrencies", "encoding"]
 keywords   = ["blockchain", "cosmos", "tendermint", "transaction"]
 
 [dependencies]
-cosmos-sdk-proto = { version = "0.3", default-features = false, path = "../cosmos-sdk-proto" }
+cosmos-sdk-proto = { default-features = false, path = "../cosmos-sdk-proto" }
```
tony-iqlusion, line 14: Without explicitly versioned dependencies, it's not possible to publish crates to https://crates.io

What problem was this causing?

tony-iqlusion (changes requested):

cosmos-tx/Cargo.toml
jkilpatr: I see, I just had to update it by hand and didn't see it as a good use of time.

Never tried to publish crates from a workspace before, usually break the crates out into their own repos. will correct this real quickly. 

Thanks!

tony-iqlusion (approved):

PR #65, opened 2021-04-02
Revert "Run `proto-build` automatically with GitHub Actions (#24)"

tony-iqlusion:
This reverts commit 95371cf40d5432194c2201037bddc90d610ec96f.

Per #62 it appears `prost-build` generates files with a somewhat nondeterministic ordering, triggering spurious PRs even when nothing is changed.

For an auto-rebuild to be helpful the build needs to be fully deterministic.


tony-iqlusion:
Closing this due to the non-determinism (see also #65)

jkilpatr (approved):
That's very annoying. Could we maybe just flag the task to run on changes to the cosmos-sdk-proto folder? I'm not sure it's worth the effort and will defer to you on that.

tony-iqlusion:
That's an interesting suggestion and might resolve some of the problems

tony-iqlusion:
Will open another PR that scopes when this action runs

PR #66, opened 2021-04-02
cosmos-tx: add SigningKey type

tony-iqlusion:
Adds a `SigningKey` type which wraps a `Box` containing a concrete
signing key implementation, allowing it to be used with either a
soft-signer (backed by the `k256` crate) or HSM signers that impl
the correct combination of traits (e.g. `yubihsm`)

This type additionally provides methods for obtaining a
`tendermint::PublicKey` for the signer, which is needed for e.g.
computing the relevant account address for the signer.

As it were, obtaining the account address is needed to implement
integration testing, as it's the first parameter we need to pass into a
single-node gaia Docker container.

### DIFF
--- a/cosmos-tx/Cargo.toml
+++ b/cosmos-tx/Cargo.toml
@@ -17,6 +17,7 @@ eyre = "0.6"
 k256 = { version = "0.7", features = ["ecdsa", "sha256"] }
 prost = "0.7"
 prost-types = "0.7"
+rand_core = "0.5"
 rust_decimal = "1.9"
-tendermint = "0.18"
+tendermint = { version = "0.18", features = ["secp256k1"] }
 thiserror = "1"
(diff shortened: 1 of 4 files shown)


jkilpatr (approved):
Looks like we're working on the same stuff right now. I'm working on upgrading [deep_space](https://github.com/althea-net/deep_spacel) with proto tx support based on cosmos-tx actaully.

Using the secp256k1 crate. Not because I like the C bindings but more because of the amount of effort that's already there in terms of deriving addresses, mnemonics and other utility functions.

I'd like to expand the trait bounds upstream here to allow for those usecases.

Anyways this is all off topic patch LGTM.

tony-iqlusion:
FWIW these same issues are coming up in [ethers-rs](https://github.com/gakonst/ethers-rs), which is using `k256`.

There should hopefully be a resolution soon.

tony-iqlusion:
> I'd like to expand the trait bounds upstream here to allow for those usecases.

Regarding this, there is a wrapper for the `secp256k1` crate which provides impls of these traits here:

https://crates.io/crates/signatory-secp256k1

However, I haven't been maintaining it as `k256` has sufficed for our needs so far.

PR #70, opened 2021-04-05
CI: constrain when rebuild-protos runs

tony-iqlusion:
Protos should only be rebuilt if files in `proto-build/` change, e.g. `COSMOS_REV` is bumped or other changes are made to the proto generator.

This should cut down on spurious rebuilds, since unfortunately the generated code does not appear to be fully deterministic.


zmanian (approved):

jkilpatr (approved):
This is more straightforward than i thought it would be.

PR #71, opened 2021-04-06
cosmos-tx: MsgSend support w\ integration test

tony-iqlusion:
Basic support for `MsgSend` with end-to-end integration test.

Adds the following:

- Traits for simplifying Protobuf serialization:
  - `prost_ext::MessageExt` for Protobuf encoding
  - `msg::MsgType` and `msg::MsgProto` for decoding/encoding `Msg` types as Protobuf `Any` messages.
- Domain types which model the following:
  - `Coin`: amount to send and a denom
  - `Denom`: name of a particular denomination
  - `Fee`: transaction fees
  - `MsgSend`: transaction message for performing a simple send

Additionally includes an end-to-end test which uses Docker to spawn a single-node `gaia` and send a transaction.

Note: at least as of the current state of this draft, the end-to-end test is not run as part of CI, but that seems possible.


tony-iqlusion:
This isn't quite working but it's getting pretty far. It's able to build a `MsgSend` transaction, spawn a single-node `gaia` network running under the `jackzampolin/gaiatest` Docker image, and uses `tendermint-rpc` to submit a transaction.

The transaction is failing validation due to a parse error, but that should be fairly trivial to fix hopefully. The good news is most of the moving pieces are talking to each other and we're down to the nitty gritty:

```
Response {
    check_tx: TxResult {
        code: Err(
            2,
        ),
        data: None,
        log: Log(
            "math/big: cannot unmarshal \"1000000.000000000000000000\" into a *big.Int: tx parse error",
        ),
        info: Info(
            "",
        ),
        gas_wanted: Gas(
            0,
        ),
        gas_used: Gas(
            0,
        ),
        events: [],
        codespace: Codespace(
            "sdk",
        ),
    },
    deliver_tx: TxResult {
        code: Ok,
        data: None,
        log: Log(
            "",
        ),
        info: Info(
            "",
        ),
        gas_wanted: Gas(
            0,
        ),
        gas_used: Gas(
            0,
        ),
        events: [],
        codespace: Codespace(
            "",
        ),
    },
    hash: transaction::Hash(1BC3F3638A43BB02766D48772D2771B29458056F2755692F369B3E34E2834124),
    height: block::Height(0),
}
```

tony-iqlusion:
One other thing I should mention:

Everything I've added in this PR is attempting to be isomorphic to `cosmos-sdk`, including the module structure and type names.

It seems the scope of types is actually pretty vast even to implement a simple send, and the more I think about it, the more this crate should probably be `cosmos-sdk` (a.k.a. "cosmos-sdk-rs" to disambiguate from the Go library, in the spirit of "tendermint-rs").

I think it'd be nice to get rid of `cosmos_tx::Builder` and replace it with a `cosmos_tx` or `cosmos_sdk::tx::TxBody` domain type, which can be signed/serialized or deserialized from protos, more or less following a similar approach to what `tendermint-rs` is doing.

I'm avoiding doing that in this particular PR to make it easier to review.

cosmos-tx/src/msg.rs
```diff
@@ -26,3 +28,48 @@ impl From<Msg> for Any {
...
+
+    /// Serialize this value as a [`Msg`].
+    fn to_msg(&self) -> Result<Msg>;
+}
+
+/// Proto types which can be used as a [`Msg`].
+pub trait MsgProto: Default + MessageExt {
+    /// Type URL value
+    const TYPE_URL: &'static str;
+}
```
tony-iqlusion, line 47: Not sure if there are other ready-made patterns for `prost_types::Any`, but this seemed reasonable enough to me given the use case.

cosmos-tx/tests/integration.rs
```diff
@@ -0,0 +1,126 @@
...
+            &[msg_send.to_msg().unwrap()],
+            fee,
+            EXAMPLE_MEMO,
+            timeout_height.into(),
+        )
+        .unwrap();
+
+    run_docker(sender_private_key.account_id(), || {
+        // TODO(tarcieri): poll for readiness instead of blindly waiting
+        thread::sleep(Duration::from_secs(2));
```
tony-iqlusion, line 63: Might be nice to have an upstream solution for this in `tendermint-rpc` (i.e. polling for readiness within a timeout waiting for a node to boot)

cc @thanethomson

jkilpatr:
Nice work @tony-iqlusion 

While you where doing this I've pulled your cosmos-tx crate into Althea's deep_space library, mostly for the same purposes.

https://github.com/althea-net/deep_space

I'm currently testing out proto tx from that codebase now. The reason I pulled code down rather than up-streaming is mostly because deep_space has a much larger scope. As a grpc library, with basic key management and mnemonic parsing and a whole pile of utility functions in addition tx signing.

Right now I have a lot of work to do to reconcile the different secp256 crates, I do some manual bit hacking for the key import and generation code that doesn't play so well with the pure rust version, which I'd rather be using in general.

I think it's worth discussing what a cosmos-sdk crate would look like. Should we be thinking of a batteries included library or series of libraries? how should they be broken out. Etc.

tony-iqlusion:
For what it's worth I've been collaborating with the [ethers-rs](https://github.com/gakonst/ethers-rs) developers on a BIP32 crate (also supporting BIP39) which is based on `k256`:

https://github.com/iqlusioninc/crates/pull/670

cosmos-tx/src/msg.rs
tony-iqlusion: Guess not, and it looks like this might all change soon: https://github.com/cosmos/cosmos-sdk/issues/9063

cosmos-tx/tests/integration.rs
tony-iqlusion: Upstreamed this and it was released in `tendermint-rpc` v0.19 🎉

tony-iqlusion:
Just pushed some changes up that are sufficient to get the signature to verify as part of a local integration test.

I'll do some additional cleanups and see if I can get the full integration test to run as part of GitHub Actions and then remove draft/WIP (hopefully tomorrow)

tony-iqlusion:
Updated the PR so it runs the `gaia` container as part of the CI workflow and the tests are passing! 🎉 

Removed draft/WIP.

cosmos-tx/src/base.rs
```diff
@@ -0,0 +1,171 @@
...
+#[derive(Clone, Eq, PartialEq, PartialOrd, Ord)]
+pub struct AccountId {
+    /// Account ID encoded as Bech32
+    bech32: String,
+
+    /// Length of the human-readable prefix of the address
+    hrp_length: usize,
+}
+
+impl AccountId {
```
jkilpatr, line 22: big thumbs on on local types for these things. 👍🏻

cosmos-tx/src/lib.rs
```diff
@@ -1,20 +1,35 @@
...
+    public_key::PublicKey,
+    signing_key::SigningKey,
 };
-pub use k256::ecdsa::{Signature, VerifyingKey};
-pub use tendermint::PublicKey;
+
+pub use k256::ecdsa::Signature;
+pub use tendermint;
+
+#[cfg(feature = "rpc")]
```
jkilpatr, line 34: I'm on the fence about the usefulness of an RPC feature, since we're pulling in Tonic as a transitive dep from cosmos-sdk-proto the only advantage an RPC feature gives is keeping us from using any tonic calls and then hoping LTO removes tonic from the final binary.

We should be able to get the same results by just not using RPC calls in a final binary, unless I'm missing somthing.

Thoughts?

cosmos-tx/tests/integration.rs
```diff
@@ -0,0 +1,191 @@
...
+
+            if tx_commit_response.check_tx.code.is_err() {
+                panic!("check_tx failed: {:?}", tx_commit_response.check_tx);
+            }
+
+            if tx_commit_response.deliver_tx.code.is_err() {
+                panic!("deliver_tx failed: {:?}", tx_commit_response.deliver_tx);
+            }
+
+            // TODO(tarcieri): look up transaction by hash and test transaction parsing
```
jkilpatr, line 125: I have the lookup grpc code, should I just upstream that as a broken out cate? I can integrate it in here.

cosmos-tx/tests/integration.rs
```diff
@@ -0,0 +1,191 @@
...
+    if let Err(err) = result {
+        panic::resume_unwind(err);
+    }
+}
+
+/// Execute a given `docker` command, returning what was written to stdout
+/// if the command completed successfully.
+///
+/// Panics if the `docker` process exits with an error code
+fn exec_docker_command<A, S>(name: &str, args: A) -> String
```
jkilpatr, line 171: this is pretty cool.

jkilpatr (approved):

cosmos-tx/tests/integration.rs
```diff
@@ -0,0 +1,191 @@
...
+            //     }
+            //
+            // This should ideally get fixed upstream, however there is
+            // presently no tracking issue for this bug.
+            //
+            // If we see it pop up locally or in CI, perhaps we could find
+            // some way of polling readiness via RPC, or retrying when this
+            // specific error is encountered.
+            // TODO(tarcieri): open upstream issue and remove this hack when fixed
+            tokio::time::sleep(Duration::from_secs(2)).await;
```
tony-iqlusion, line 113: This is probably the most suspect part of this PR.

It seems like polling an RPC endpoint to determine when the first block has been produced might help.

cosmos-tx/tests/integration.rs
tony-iqlusion, line 125: I'd like to do this, but this PR is already so large I thought it was worth skipping for now and doing it in a follow-up.

cosmos-tx/src/lib.rs
tony-iqlusion: We need it for the tests anyway, so it can either exist as test-specific functionality or a reusable part of the public API. I opted for the latter.

It's an off-by-default optional feature, so it doesn't seem problematic to me for users who aren't interested in it.

cosmos-tx/tests/integration.rs
tony-iqlusion: Since the tests are already using RPC, I can just hit the `/block` endpoint:

https://docs.rs/tendermint-rpc/0.19.0/tendermint_rpc/trait.Client.html#method.latest_block

tony-iqlusion: Since we already have an RPC client instantiated, I can use the `/tx_search` endpoint to look it up by hash:

https://docs.rs/tendermint-rpc/0.19.0/tendermint_rpc/trait.Client.html#method.tx_search

cosmos-tx/tests/integration.rs
tony-iqlusion: I was thinking of extracting this functionality in such a way it's reusable, e.g. behind a `dev` feature.

I think it'd be useful for testing TMKMS transaction signing.

But I thought I'd save that for a follow-up PR.

cosmos-tx/tests/integration.rs
tony-iqlusion: Replaced this `sleep` with a `wait_for_first_block` function that polls the `/latest_block` endpoint and waits for a valid result.

cosmos-tx/src/lib.rs
jkilpatr: Def agree that it should be a part of the public API. I just don't think feature gating it is useful if we don't reduce the dependency tree by doing so.

I don't have very strong feelings about this, just a comment.

tony-iqlusion: `cosmos-sdk-proto` is being pulled in with `default-features = false`, so it isn't pulling in Tonic:

https://github.com/cosmos/cosmos-rust/blob/main/cosmos-tx/Cargo.toml#L14

PR #73, opened 2021-04-08
Add imports for secp256k1 and ed25519 proto defs

jkilpatr:
I missed these in my previous comments, just a few minor key type
structs.


tony-iqlusion (approved):

tony-iqlusion:
Hmmm, curious if these might be the cause of the error I've been encountering...

PR #76, opened 2021-04-12
cosmos-tx: transaction domain types and parser

tony-iqlusion:
Adds "rustic" domain types for transactions, including:

- `Tx`: transactions
- `AuthInfo`: fee and signer mode info
- `ModeInfo`: single versus multisig
- `SignDoc`: transaction document to be signed
- `SignerInfo`: information about signers

Additionally extracts a `crypto` module which contains public keys, signing keys, and `CompactBitArray` used to represent the signers of a multisignature transaction.

These are then used in the integration test to parse the resulting transaction and ensure it's the one we constructed.


tony-iqlusion:
Note: the eventual goal of this is to replace the current `cosmos_tx::Builder` type with these, which would also make it possible to express multisig transactions using the public API.

I can either leave that work as a followup to make this PR easier to review, or add it in for completeness depending on what people would prefer.

cosmos-tx/src/crypto/compact_bit_array.rs
```diff
@@ -0,0 +1,40 @@
...
+
+use cosmos_sdk_proto::cosmos;
+
+/// [`CompactBitArray`] is an implementation of a space efficient bit array.
+///
+/// This is used to ensure that the encoded data takes up a minimal amount of
+/// space after proto encoding.
+#[derive(Clone, Debug, PartialEq)]
+pub struct CompactBitArray {
+    // TODO(tarcieri): better internal representation for this, e.g. `bitvec`
```
tony-iqlusion, line 11: Thought I'd postpone making this decision for now, although I'd probably recommend `bitvec`.

cosmos-tx/tests/integration.rs
```diff
@@ -99,9 +99,26 @@ fn msg_send() {
                 panic!("deliver_tx failed: {:?}", tx_commit_response.deliver_tx);
             }
 
-            // TODO(tarcieri): look up transaction by hash and test transaction parsing
-        });
+            // TODO(tarcieri): avoid race here between when API reports TX available
+            time::sleep(Duration::from_secs(1)).await;
```
tony-iqlusion, line 103: It seems as though even if `DeliverTx` has succeeded the transaction is sporadically missing from `/tx_search`. Perhaps we need to wait until the next block here?

cosmos-tx/src/tx/signer_info.rs
```diff
@@ -0,0 +1,72 @@
...
+
+    /// Get the [`AuthInfo`] for this signer with the given fee.
+    ///
+    /// This is primarily useful for cases involving a single signer.
+    pub fn auth_info(self, fee: Fee) -> AuthInfo {
+        AuthInfo {
+            signer_infos: vec![self],
+            fee,
+        }
+    }
```
tony-iqlusion, line 46: Added these as convenience methods for the single direct signer use case, which seems common/important.

The intended usage looks like:

```rust
SignerInfo::single_direct(Some(sender_public_key), sequence_number).auth_info(fee)
```

It's also used in the tests now.

tony-iqlusion:
Went ahead and completely replaced the `cosmos_tx::Builder` type with the domain types, namely by adding a `SignDoc` type.

Going to investigate a better solution to the `/tx_search` race condition, but otherwise I'd call this done.

jkilpatr (approved):
Looks good to me, one question, where are the test cases for the parser? I didn't see any

tony-iqlusion:
It tests the transaction round trips successfully in the integration test by looking it up via RPC, parsing it, and then comparing the body and auth info match what was originally submitted:

https://github.com/cosmos/cosmos-rust/pull/76/files#diff-0e31f23fc223d1d543169a682e0f1093c1cc53c6a312febce7d5d1854919fafaR102-R104

marbar3778 (commented):
it might make sense to touch on signature support. This only supports secp correct?

tony-iqlusion:
@marbar3778 for now, yes.

I've factored the relevant types into `crypto::secp256k1` in preparation for eventually supporting more algorithms (e.g. ECDSA/P-256).

It also uses a newtype of `tendermint::PublicKey`, so if additional algorithms get added upstream in tendermint-rs, it will be easy to add support for them.

PR #77, opened 2021-04-13
cosmos-sdk-rs: add initial usage docs and `dev` module

tony-iqlusion:
Adds basic documentation for building, signing, and parsing transactions.

Update: went ahead and extracted some of the functions used in the integration test into a `dev` module with the goal of reusing them in downstream consumers of this crate.

TMKMS and Delphi are good candidates for the `dev` module. I imagine there are other use cases that might make sense (e.g. Gravity Orchestrator).

It might make sense to upstream some of that to tendermint-rs? /cc @thanethomson


tony-iqlusion:
Going to go ahead and land this, but wait for feedback on a final release PR.

If anyone has any issues with this, I can address them before cutting a final release.

PR #79, opened 2021-04-14
Fix links to `cosmos_sdk`

tony-iqlusion:
Unfortunately it was originally published with an underscore instead of a hyphen, and it's not presently possible to change that.

Hyphens work in place of an underscore in most places, however one place they don't, apparently is the image for docs.rs badges (it does work for links, however).

This commit fixes the badges, uses `cosmos_sdk` in links for consistency with the current name, and linkifies badges in the table in the toplevel README.md.


jkilpatr (approved):
LGTM

PR #82, opened 2021-05-14
Add staking functions to cosmos-sdk

kimonp:
Add staking functions to the cosmos-sdk by extending to existing protobuf methods:

- MsgDelegate
- MsgUnDelegate
- MsgBeginRedelegate

### DIFF
--- a/cosmos-sdk-rs/src/staking.rs
+++ b/cosmos-sdk-rs/src/staking.rs
@@ -166,10 +166,10 @@ pub struct MsgBeginRedelegate {
     pub delegator_address: AccountId,
 
     /// Source validator's address.
-    pub src_validator_address: AccountId,
+    pub validator_src_address: AccountId,
 
     /// Destination validator's address.
-    pub dst_validator_address: AccountId,
+    pub validator_dst_address: AccountId,
 
     /// Amount to UnDelegate
     pub amount: Option<Coin>,
@@ -207,8 +207,8 @@ impl TryFrom<&proto::cosmos::staking::v1beta1::MsgBeginRedelegate> for MsgBeginR
         };
         Ok(MsgBeginRedelegate {
             delegator_address: proto.delegator_address.parse()?,
-            src_validator_address: proto.src_validator_address.parse()?,
-            dst_validator_address: proto.dst_validator_address.parse()?,
+            validator_src_address: proto.validator_src_address.parse()?,
+            validator_dst_address: proto.validator_dst_address.parse()?,
             amount,
         })
     }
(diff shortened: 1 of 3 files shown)


kimonp:
@tony-iqlusion , @jkilpatr : this pull request adds three staking functions to the cosmos-sdk by extending to existing protobuf methods.  Hopefully others will find this useful as well, please take a look.

cosmos-sdk-rs/src/staking.rs
```diff
@@ -0,0 +1,240 @@
...
+#[derive(Clone, Debug, Eq, PartialEq, PartialOrd, Ord)]
+pub struct MsgDelegate {
+    /// Delegator's address.
+    pub delegator_address: AccountId,
+
+    /// Validator's address.
+    pub validator_address: AccountId,
+
+    /// Amount to send
+    pub amount: Option<Coin>,
```
tony-iqlusion, line 22: I'm curious... when would this ever be `None`?

tony-iqlusion (approved):
Sorry this took so long to review! LGTM

tony-iqlusion:
@kimonp this just needs a rustfmt to be mergeable, but also there is one open question it'd be good to get an answer to

tony-iqlusion:
Going to merge this. I'll correct the rustfmt issue locally (I'm opening another PR anyway)

cosmos-sdk-rs/src/staking.rs
kimonp: proto::cosmos::staking::v1beta1::MsgDelegate's amount can be None, so I'm mirroring that here.

tony-iqlusion: The Rust code generated from the protos has no way to express non-optional fields for messages, so that's not a great thing to go off of.

Looking at the raw Protobuf definition,  it's:

https://github.com/cosmos/cosmos-sdk/blob/9fd866e3820b3510010ae172b682d71594cd8c14/proto/cosmos/staking/v1beta1/tx.proto#L89

```proto
message MsgDelegate {
  option (gogoproto.equal)           = false;
  option (gogoproto.goproto_getters) = false;

  string                   delegator_address = 1 [(gogoproto.moretags) = "yaml:\"delegator_address\""];
  string                   validator_address = 2 [(gogoproto.moretags) = "yaml:\"validator_address\""];
  cosmos.base.v1beta1.Coin amount            = 3 [(gogoproto.nullable) = false];
}
```

It's annotated as non-nullable (using a custom gogoproto directive which the Rust proto parser doesn't understand), so this should be changed to `pub amount: Coin` and messages missing this field should be rejected.

PR #85, opened 2021-06-10
Add more detail to AccountId parsing errors

### LINKED ISSUE #84: Parsing errors lack detail
We're building an app that needs to decode a `tx::Tx` that contains a `bank::MsgSend` message, and when we supply an invalid `from_address` or `to_address` the error produced is simply:

```
bad encoding
```

This comes from the `subtle-encoding` crate over [here](https://github.com/iqlusioninc/crates/blob/1d70c025a67f30ce893149551cc6f93b73f363d9/subtle-encoding/src/error.rs#L36) (it took us a while to realize it came from here specifically).

It'd be super helpful to have more descriptive errors when decoding these data structures and converting them into their domain types.

thanethomson:
Closes #84

### DIFF
--- a/cosmos-sdk-rs/src/base.rs
+++ b/cosmos-sdk-rs/src/base.rs
@@ -31,7 +31,11 @@ impl AccountId {
                 hrp_length: prefix.len(),
             })
         } else {
-            Err(Error::AccountId { id }.into())
+            Err(Error::AccountId {
+                id,
+                detail: "expected prefix to be lowercase alphabetical characters only".to_owned(),
+            }
+            .into())
         }
     }
 
@@ -71,7 +75,10 @@ impl FromStr for AccountId {
     type Err = eyre::Report;
 
     fn from_str(s: &str) -> Result<Self> {
-        let (hrp, bytes) = bech32::decode(s)?;
+        let (hrp, bytes) = bech32::decode(s).map_err(|e| Error::AccountId {
+            id: s.to_owned(),
+            detail: format!("failed to decode bech32: {}", e.to_string()),
+        })?;
 
         if bytes.len() == tendermint::account::LENGTH {
             Ok(Self {
@@ -79,7 +86,15 @@ impl FromStr for AccountId {
                 hrp_length: hrp.len(),
             })
         } else {
-            Err(Error::AccountId { id: s.to_owned() }.into())
+            Err(Error::AccountId {
+                id: s.to_owned(),
+                detail: format!(
+                    "account ID should be at least {} bytes long, but was {} bytes long",
+                    tendermint::account::LENGTH,
+                    bytes.len()
+                ),
+            }
+            .into())
         }
     }
 }
--- a/cosmos-sdk-rs/src/error.rs
+++ b/cosmos-sdk-rs/src/error.rs
@@ -9,10 +9,12 @@ use thiserror::Error;
 #[derive(Clone, Debug, Error, PartialEq)]
 pub enum Error {
     /// Invalid account.
-    #[error("invalid account ID: {id:?}")]
+    #[error("invalid account ID: {id:?} ({detail:?})")]
     AccountId {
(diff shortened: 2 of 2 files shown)


cosmos-sdk-rs/src/base.rs
```diff
@@ -71,15 +75,26 @@ impl FromStr for AccountId {
...
-            Err(Error::AccountId { id: s.to_owned() }.into())
+            Err(Error::AccountId {
+                id: s.to_owned(),
+                detail: format!(
+                    "account ID should be at least {} bytes long, but was {} bytes long",
+                    tendermint::account::LENGTH,
+                    bytes.len()
+                ),
+            }
+            .into())
```
tony-iqlusion, line 97: FYI, since we're using `eyre`, it already provides the functionality to attach arbitrary messages like this to errors without requiring each variant carry along a `String` field:

```suggestion
use eyre::WrapErr;

Err(Error::Address).wrap_err_with(|| {
    format!(
        "account ID should be at least {} bytes long, but was {} bytes long",
        tendermint::account::LENGTH,
        bytes.len()
    ),
})
```

tony-iqlusion (changes requested):
I think these error messages are great to have, but I'd suggest using the `eyre::WrapErr` trait (which is impl'd for `Result`) as the mechanism for implementing them.

cosmos-sdk-rs/src/error.rs
```diff
@@ -9,10 +9,12 @@ use thiserror::Error;
 #[derive(Clone, Debug, Error, PartialEq)]
 pub enum Error {
     /// Invalid account.
-    #[error("invalid account ID: {id:?}")]
+    #[error("invalid account ID: {id:?} ({detail:?})")]
     AccountId {
         /// Malformed account ID
         id: String,
+        /// Specific detail as to why this account ID is malformed.
+        detail: String,
```
tony-iqlusion, line 17: If you use `eyre::WrapErr`, this field is unnecessary.

cosmos-sdk-rs/src/base.rs
thanethomson: Done!

cosmos-sdk-rs/src/base.rs
```diff
@@ -31,7 +32,8 @@ impl AccountId {
                 hrp_length: prefix.len(),
             })
         } else {
-            Err(Error::AccountId { id }.into())
+            Err::<Self, eyre::Report>(Error::AccountId { id }.into())
```
tony-iqlusion, line 35: Haha wow, turbofishing the `Result::Err` tuple struct variant constructor... 🤯

I'm curious though... were you hitting some sort of type inference issue otherwise? I'm a bit unclear why this is necessary...

tony-iqlusion (approved):

cosmos-sdk-rs/src/base.rs
thanethomson: Yeah I'm sure there are other ways of providing type info to the compiler, but this seemed like the least amount of code :slightly_smiling_face: 

The error I was getting without explicitly indicating types while using `wrap_err`:

```
error[E0282]: type annotations needed
  --> cosmos-sdk-rs/src/base.rs:35:13
   |
35 |             Err(Error::AccountId { id }.into())
   |             ^^^ ------------------------------ this method call resolves to `T`
   |             |
   |             cannot infer type for type parameter `E` declared on the enum `Result`

error: aborting due to previous error
```

It's unfortunate, since I'd assume, given that the function result type is `eyre::Result<Self>`, that the compiler would be smart enough to figure this out for itself :disappointed:

tony-iqlusion: Aah, I think it was the `.into()`s. You don't need them with `wrap_err`, and since it had them it didn't know what type to infer.

Will open a small cleanup PR.

PR #88, opened 2021-06-22
WASM support with grpc-web

shravanshetty1:
Regenerated files by disabling the "transport" feature in tonic, this will allow "cosmos-sdk-proto" to be imported into wasm based applications.


tony-iqlusion:
We already support WASM and test it in CI:

https://github.com/cosmos/cosmos-rust/blob/main/.github/workflows/cosmos-sdk-proto.yml#L20-L38

Perhaps we need to better document it.

To support WASM, add `default-features = "false"` to the `cosmos-sdk-proto` requirement in your `Cargo.toml`:

```toml
[dependencies.cosmos-sdk-proto]
version = "0.5"
default-features = false
```

shravanshetty1:
> We already support WASM and test it in CI:
> 
> https://github.com/cosmos/cosmos-rust/blob/main/.github/workflows/cosmos-sdk-proto.yml#L20-L38
> 
> Perhaps we need to better document it.
> 
> To support WASM, add `default-features = "false"` to the `cosmos-sdk-proto` requirement in your `Cargo.toml`:
> 
> ```toml
> [dependencies.cosmos-sdk-proto]
> version = "0.5"
> default-features = false
> ```

Yes but this will disable grpc, this pr is to support grpc-web - which requires tonic

shravanshetty1:
People can use https://github.com/titanous/grpc-web-client along with this to support grpc-web in wasm. Eventually tonic will have this builtin in some way which is tracked in this issue https://github.com/hyperium/tonic/issues/645

tony-iqlusion:
Aha! In that case, can you remove the `grpc` feature, as `grpc` will otherwise be always available (and your code changes have removed the feature gating anyway)

tony-iqlusion:
@shravanshetty1 mind rebasing?

shravanshetty1:
https://github.com/cosmos/cosmos-rust/pull/90

PR #90, opened 2021-06-23
[WIP] WASM support with grpc-web

shravanshetty1:
@tony-iqlusion messed up the rebase

### DIFF
--- a/cosmos-sdk-proto/Cargo.toml
+++ b/cosmos-sdk-proto/Cargo.toml
@@ -17,10 +17,14 @@ keywords   = ["blockchain", "cosmos", "tendermint", "proto"]
 [dependencies]
 prost = "0.7"
 prost-types = "0.7"
-tonic = { version = "0.4.3", default-features = false, features = ["codegen", "prost"] }
+tonic = { version = "0.4.3", optional = true, default-features = false, features = ["codegen", "prost"] }
 
 [dependencies.tendermint-proto]
 version = "0.20"
+
+[features]
+default = ["grpc"]
+grpc = ["tonic"]
 
 [package.metadata.docs.rs]
 all-features = true
--- a/cosmos-sdk-proto/Cargo.toml
+++ b/cosmos-sdk-proto/Cargo.toml
@@ -17,16 +17,10 @@ keywords   = ["blockchain", "cosmos", "tendermint", "proto"]
 [dependencies]
 prost = "0.7"
 prost-types = "0.7"
-
-# Optional dependencies
-tonic = { version = "0.4", optional = true }
+tonic = { version = "0.4.3", default-features = false, features = ["codegen", "prost"] }
 
 [dependencies.tendermint-proto]
 version = "0.20"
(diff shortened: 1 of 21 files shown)


shravanshetty1:
@tony-iqlusion Cannot remove the grpc feature since it conditionally compiles some grpc related code, which is needed for grpc_web. I think this breaks normal grpc clients and only supports grpc_web. Please review

tony-iqlusion:
@shahankhatch looks good. want to remove `[WIP]`?

tony-iqlusion (approved):

shravanshetty1:
> @shahankhatch looks good. want to remove `[WIP]`?

Unfortunately after looking more closely at the code thats being deleted. I can confidently say that this will only support grpc_web, normal grpc wont work. Will test it out when i get some time.

I am not sure whats the best solution here, we could make a seperate directoy called "cosmos-sdk-proto-wasm" or another solution could be to wait for "tonic" to enable grpc_web using some kind of feature which will be tracked in issue https://github.com/hyperium/tonic/issues/645. People who want wasm grpc-web till than can use my fork.

tony-iqlusion:
@shravanshetty1 you can conditionally gate how the `tonic` dependency is included depending on whether or not it's a WASM target:

https://github.com/cosmos/cosmos-rust/blob/main/cosmos-sdk-rs/Cargo.toml#L27-L28

shravanshetty1:
> @shravanshetty1 you can conditionally gate how the `tonic` dependency is included depending on whether or not it's a WASM target:
> 
> https://github.com/cosmos/cosmos-rust/blob/main/cosmos-sdk-rs/Cargo.toml#L27-L28

By disabling tonic, we can get wasm compatibility but we cant use grpc_web. To use grpc_web we will have to not disable tonic, but only disable the transport feature of tonic. This will make the generated code wasm compatible and allow us to use grpc_web using clients such as https://github.com/titanous/grpc-web-client.

However, since the generated code without transport feature of tonic wont allow us to use normal grpc and only force us to use grpc_web.  

We could generate 2 cosmos-sdk-proto, one which is wasm+grpc_web(cosmos-sdk-proto-wasm) compatible and the other is the existing one. 
We could also just wait for tonic to support grpc_web. After they support it the tonic generated code will have compilation flags which will allow us to compile for grpc_web wasm and normal based on situation.

Its your call.

tony-iqlusion:
Aah, I guess this is where we're running into drawbacks with having pre-generated/checked-in code, in that we can't specialize things to a particular platform.

I guess for now if you have a working fork that's fine, and we should perhaps circle back on this issue after the next release (it'd be good to get one out ASAP as there's a new version of tendermint-rs)

shravanshetty1:
For any future readers - you can find wasm+grpc_web compatible version https://github.com/shravanshetty1/cosmos-rust/tree/wasm2

shravanshetty1:
Actually after manual testing works fine for normal grpc, not sure how

PR #91, opened 2021-06-24
cosmos-sdk-rs: minimal `bip32` integration

tony-iqlusion:
Adds basic support for deriving a `crypto::secp256k1::SigningKey` using a provided BIP32 derivation path.


jkilpatr (approved):
This looks good to me. Can we flesh out this trait definition any more while we're at it? Having a trait that gives the signing key seems less complete than returning a signature of some kind (allowing the implementation to handle signing internally). Or do you have a different architecture in mind for diverse key support?

This is a great starting point either way.

tony-iqlusion:
> Can we flesh out this trait definition any more while we're at it? Having a trait that gives the signing key seems less complete than returning a signature of some kind

It does, although perhaps that could be better documented:

```rust
pub trait EcdsaSigner: ecdsa::signature::Signer<Signature> {
    /// Get the ECDSA verifying key for this signer
    fn verifying_key(&self) -> VerifyingKey;
}
```

Notably it's bounded on `ecdsa::signature::Signer`, which is a re-export of this trait:

https://docs.rs/signature/1.3.0/signature/trait.Signer.html

```rust
pub trait Signer<S: Signature> {
    pub fn try_sign(&self, msg: &[u8]) -> Result<S, Error>;

    pub fn sign(&self, msg: &[u8]) -> S { ... }
}
```

This allows you to plug in whatever signing implementation you like, so long as it returns an ECDSA/secp256k1 signature.

For example, the `yubihsm` crate provides an implementation of this trait backed by an HSM signer:

https://docs.rs/yubihsm/0.39.0/yubihsm/ecdsa/secp256k1/type.Signer.html

tony-iqlusion:
I pushed up some documentation which hopefully better describes how the trait works and how you can use an alternative [`EcdsaSigner`] implementation.

tony-iqlusion:
@jkilpatr hopefully those doc improvements work. gonna merge this and make a release PR, and I can tackle any remaining changes in that

PR #96, opened 2021-07-01
Add basic `cosmwasm` support

jstuczyn:
While using the library in conjunction with [cosmwasm](https://github.com/CosmWasm/wasmd)-based cosmos validators, I found myself writing bunch of code that I figured could instead be contributed back to the community. 

This pull request introduces basic support for `wasmd`. It includes compilation of the protobuf files and conversion of Msg types into something that can be more easily put into Tx Body.

I realise most people would not need it, so I've left it feature-locked behind `cosmwasm` feature flag.


tony-iqlusion:
Thanks! This is definitely interesting.

I'm on vacation and don't have time to review it right now, but perhaps @jkilpatr could take a look in the meantime

zmanian:
This looks good to me.

I don't really know what the answer should be for tracking compatibility with multiple upstream versions should be.

wasmd and the cosmos SDK being versioned independently.

tony-iqlusion (approved):
Looks good, thanks!

PR #108, opened 2021-07-30
Update protobufs for cosmos-sdk 0.43.0-rc2

zmanian:


tony-iqlusion (approved):

tony-iqlusion:
FYI, protos will be regenerated automatically, so all that's needed is the version bump.

Not sure how @jkilpatr or others feel in particular about me merging @zmanian's PR since we're both at @iqlusioninc, but hopefully this is innocuous enough, and the protos will be regened via `proto-build` anyway. Let me know if you think it's a problem.

Note: we should also bump tendermint-rs before another release.

PR #115, opened 2021-08-04
Made admin field optional in MsgInstantiateContract

jstuczyn:
I've run into my own omission from the previous pull request - as the comment suggests, the admin field should be optional.

### DIFF
--- a/cosmos-sdk-rs/src/cosmwasm.rs
+++ b/cosmos-sdk-rs/src/cosmwasm.rs
@@ -174,9 +174,14 @@ impl TryFrom<proto::cosmwasm::wasm::v1beta1::MsgInstantiateContract> for MsgInst
         } else {
             Some(proto.label)
         };
+        let admin = if proto.admin.is_empty() {
+            None
+        } else {
+            Some(proto.admin.parse())
+        };
         Ok(MsgInstantiateContract {
             sender: proto.sender.parse()?,
-            admin: proto.admin.map(|admin| admin.parse()).transpose()?,
+            admin: admin.transpose()?,
             code_id: proto.code_id,
             label,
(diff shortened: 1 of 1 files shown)


cosmos-sdk-rs/src/cosmwasm.rs
```diff
@@ -174,9 +174,14 @@ impl TryFrom<proto::cosmwasm::wasm::v1beta1::MsgInstantiateContract> for MsgInst
         } else {
             Some(proto.label)
         };
+        let admin = if proto.admin.is_empty() {
+            None
+        } else {
+            Some(proto.admin.parse())
+        };
```
tony-iqlusion, line 181: Any reason you didn't write this as...

```suggestion
        let admin = proto.admin.map(|admin| admin.parse());
```

jstuczyn: Honestly, didn't think of it at the time, good catch! I've applied the suggestion in the next commit and moved it directly to field assignment to keep it all more compact.

tony-iqlusion: Oh whoops, now I see why: it's a `String`, and you do want the original behavior. My bad.

Unfortunately we don't really have a good helper for that either. The closest thing is probably [` tendermint_proto::serializers::optional_from_str`](https://docs.rs/tendermint-proto/0.21.0/tendermint_proto/serializers/index.html), although I wouldn't suggest using that.

Mind reverting? Sorry again.

jstuczyn: Absolutely. I'm also sorry for not properly checking the suggestion myself. I forgot to pass `features=cosmwasm` and assumed it was all fine...

PR #116, opened 2021-08-06
cosmos-sdk-rs: make `Tx::find_by_hash` use the `/tx` endpoint

tony-iqlusion:
Notably this endpoint provides read-your-writes consistency with regard to transactions, whereas `/tx_search` does not.

See also: tendermint/tendermint#6359


tony-iqlusion:
This is now working locally for me now that https://github.com/informalsystems/tendermint-rs/pull/948 was fixed upstream in tendermint-rs v0.22

tony-iqlusion:
@jkilpatr can I get a stamp for this?

Unfortunately I wasn't able to eliminate the need to poll for the transaction in the tests. Perhaps there's a lingering issue with https://github.com/tendermint/tendermint/issues/6359

However, the `/tx` endpoint still seems more appropriate here than `/tx_search`

zmanian (approved):

PR #119, opened 2021-08-24
New protos for vesting accounts

amromashov:
**Changed**

- Rebuild protos with cosmos-sdk v0.43.0 (added "proto/cosmos/vesting/v1beta1/vesting.proto" to proto paths)


tony-iqlusion:
Thank you!

Issue #113, opened 2021-08-03
Rename `cosmos_sdk` again: `cosmrs`?


tony-iqlusion:
After we got the `cosmos_sdk` crate, I was somewhat disappointed to discover that it's locked into using an underscore, which is somewhat annoying given it doesn't match `cosmos-sdk-proto`.

Additionally, the `cosmos-sdk-rs` branding isn't great in many ways: it implements a minute amount of functionality of the actual Go `cosmos-sdk`. Given that I think it inevitably sets a user up for disappointment when they're looking for a full-blown analog of the upstream Go library.

I've been wondering if a pithier name for a "Cosmos Rust Support Library" might be appropriate. The name I've been batting around in my head is "cosmrs", as in cosmos + rs. I think it also pairs nicely with CosmWasm.

WDYT?

ethanfrey:
I like it. It also pairs nicely with CosmJS, which is a client side js lib.

Cosmrs or cosm-rs, both work for me. But the name seems more clear (I was expecting a server side sdk as well)

adizere:
Clarifying that the repo is a "Cosmos Rust Support Library" would definitely improve usability and expectations.

andynog:
I would rather it to be called `cosm-rs` to match other Rust projects naming like  `tendermint-rs` and `ibc-rs`. Also, clarifying that this is a client side library would be important.

Another thing that I think it'd be important for the ecosystem would be to eventually track the client side implementations features so users can understand what is already implemented or not, for example CosmJS currently probably offers the broadest implementation support for modules (bank, auth, etc). This could be useful for other client libraries too like Alan.dart (Dart) for example.

tony-iqlusion:
> I would rather it to be called cosm-rs to match other Rust projects naming like tendermint-rs and ibc-rs. 

We've been tacking `-rs` onto the name (cosmos-sdk-rs) even though the crate is named `cosmos_sdk`, in the same way that tendermint-rs is technically the `tendermint` crate and ibc-rs is technically the `ibc` crate.

However, in this case I'm proposing actually putting it in the crate name and calling it `cosmrs`, which is more of a play on changing `o` -> `r`.

tony-iqlusion:
Renamed to `cosmrs` in #125, with v0.1.0 released in #127

PR #125, opened 2021-08-25
Rename `cosmos-sdk-rs` to `cosmrs`

tony-iqlusion:
As discussed in #113, this renames "cosmos-sdk-rs" (i.e. the `cosmos_sdk` crate) to `cosmrs`.


romac:
What do you think about dropping the `rs` suffix and naming the crate just `cosm`? (which you incidentally seem to own on crates.io already)

We can keep the suffix for the repo.

ethanfrey (approved):
Looks good to me.

`cosm` would work too *as a crate name* but when referring to it in the ecosystem at large, it is nice to have some language reference. eg. cosmjs (even though everything on npm is clearly js) I would stick with `cosmrs`

tony-iqlusion:
Just as a point of comparison, the CosmJS NPM packages are all within the `@cosmjs` namespace.

Whereas "tendermint-rs" is published under the `tendermint` crate, and "ibc-rs" under the `ibc` crate, I don't feel like `cosm` is particularly meaningful as a crate name in and of itself.

If the `cosmos` crate weren't already taken, I think it'd be great to have "cosmos-rs" with the `cosmos` crate, but alas, it was taken long ago for a project not related to the Cosmos Network.

So I think calling the crate `cosmrs` is probably the least surprising thing if the project is named CosmRS.

PR #126, opened 2021-08-25
Bumped version of cosmos-sdk-proto in cosmos_sdk

amromashov:
cosmos_sdk still uses cosmos-sdk-proto v0.6


tony-iqlusion:
Oops, saw this after merging #125.

Also this isn't strictly necessary: a `0.6` expression in Cargo.toml is actually a SemVer `^0.6` expression, which expands to `>= 0.6.0`.

So it will always use the latest version, and in a project you just need to run `cargo update -p cosmos-sdk-proto` to update that particular package in your Cargo.lock.

Gonna go ahead and close this, but I'm about to cut an initial release of `cosmrs` under the new name, and will go ahead and pin to at least 0.6.3 for that.

PR #128, opened 2021-09-23
Update to tendermint-rs v0.22 crates

tony-iqlusion:
Release notes:

https://github.com/informalsystems/tendermint-rs/blob/master/CHANGELOG.md#v0220


jkilpatr (approved):
LTM. What's required before the -pre is removed?

tony-iqlusion:
@jkilpatr I might look at doing one other small refactoring PR first, but other than that it's good to go

PR #129, opened 2021-09-27
cosmos-sdk-proto v0.7.0

tony-iqlusion:
### Changed
- Update `tendermint` crate to v0.22 ([#128])

[#128]: https://github.com/cosmos/cosmos-rust/pull/128


tony-iqlusion:
Note: I took a look at bumping `COSMOS_REV` to v0.40.0 but it did not appear to have any effect on the generated protos.

I can still do that for housekeeping purposes if anyone is interested.

jkilpatr:
I think we should probably bump for housekeeping purpoes, just so an observer knows it's 'up to date' even if it has no actual changes.

jkilpatr (approved):

PR #133, opened 2021-09-29
cosmrs: `PublicKey` JSON serialization support

tony-iqlusion:
Adds support for the "Cosmos JSON" `PublicKey` serialization, e.g.:

    {"@type":"/cosmos.crypto.ed25519.PubKey","key":"sEEsVGkXvyewKLWMJbHVDRkBoerW0IIwmj1rHkabtHU="}

Encoding follows Protobub JSON conventions, with binary data encoded as standard (i.e. non-URL safe) Base64. However, note that it's structurally still a bit different from the Protobuf JSON encoding for public keys, and thus entails a custom solution.

Also note that there is unfortunately not yet a general solution for Protobuf JSON encoding for `prost::Message`:

https://github.com/tokio-rs/prost/issues/75

### DIFF
--- a/cosmrs/Cargo.toml
+++ b/cosmrs/Cargo.toml
@@ -18,6 +18,8 @@ k256 = { version = "0.9", features = ["ecdsa", "sha256"] }
 prost = "0.7"
 prost-types = "0.7"
 rand_core = { version = "0.6", features = ["std"] }
+serde = { version = "1", features = ["serde_derive"] }
+serde_json = "1"
 subtle-encoding = { version = "0.5", features = ["bech32-preview"] }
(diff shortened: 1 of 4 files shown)


cosmrs/Cargo.toml
```diff
@@ -18,6 +18,8 @@ k256 = { version = "0.9", features = ["ecdsa", "sha256"] }
 prost = "0.7"
 prost-types = "0.7"
 rand_core = { version = "0.6", features = ["std"] }
+serde = { version = "1", features = ["serde_derive"] }
+serde_json = "1"
```
tony-iqlusion, line 22: These are already transitive hard dependencies, so no point in making them optional

zmanian (approved):

PR #136, opened 2021-10-06
Christianborst/update sdk v0.44.1

ChristianBorst:


ChristianBorst:
Found an issue where the ibc module having its own repo means cosmos-sdk-proto generates no files

PR #157, opened 2021-11-30
GRPC web compatibility

shravanshetty1:
Works after testing

### DIFF
--- a/cosmos-sdk-proto/src/prost/cosmos.auth.v1beta1.rs
+++ b/cosmos-sdk-proto/src/prost/cosmos.auth.v1beta1.rs
@@ -88,17 +88,6 @@ pub mod query_client {
     pub struct QueryClient<T> {
         inner: tonic::client::Grpc<T>,
     }
-    impl QueryClient<tonic::transport::Channel> {
-        #[doc = r" Attempt to create a new client by connecting to a given endpoint."]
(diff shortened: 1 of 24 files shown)


cosmos-sdk-proto/src/prost/cosmos.auth.v1beta1.rs
```diff
@@ -88,17 +88,6 @@ pub mod query_client {
...
-        #[doc = r" Attempt to create a new client by connecting to a given endpoint."]
-        pub async fn connect<D>(dst: D) -> Result<Self, tonic::transport::Error>
-        where
-            D: std::convert::TryInto<tonic::transport::Endpoint>,
-            D::Error: Into<StdError>,
-        {
-            let conn = tonic::transport::Endpoint::new(dst)?.connect().await?;
-            Ok(Self::new(conn))
-        }
-    }
```
tony-iqlusion, line 101: This seems somewhat problematic...

shravanshetty1: Anyway I can merge this? - I can close it if this is not going to happen

tony-iqlusion: I'm happy to merge this if it doesn't cause a regression for non-WASM use cases, but this appears to be a blocker

PR #161, opened 2021-12-13
Pin tendermint-rs deps to v0.23.1 for now

thanethomson:
As per the tendermint-rs versioning (https://github.com/informalsystems/tendermint-rs#versioning), we're now focusing on the v0.24.0 release for compatibility with Tendermint v0.35.x, so any further breaking changes to the v0.23.x series of tendermint-rs need to be released as patch versions.

This pins your tendermint-rs deps to v0.23.1 for now until we can ensure that v0.23.2 (https://github.com/informalsystems/tendermint-rs/pull/1043) does not break anything.


tony-iqlusion:
Thank you!

Issue #174, opened 2022-01-26
Question: Should field `amount` be None in MsgUndelegate


stevenatcrypto:
Hi, I found the field `amount` in MsgUndelete is type of `Option<Coin>` (different in MsgDelegate). Could it be assigned to None? And why? Thanks.

https://github.com/cosmos/cosmos-rust/blob/main/cosmrs/src/staking.rs#L83

tony-iqlusion:
It's [non-nullable in the schema](https://github.com/cosmos/cosmos-sdk/blob/9fd866e3820b3510010ae172b682d71594cd8c14/proto/cosmos/staking/v1beta1/tx.proto#L120) and should probably receive the same treatment as #124

PR #175, opened 2022-01-27
Remove Option from field amount of MsgUndelegate and MsgBeginRedelegate

stevenatcrypto:
Close https://github.com/cosmos/cosmos-rust/issues/174

According to [MsgUndelegate](https://github.com/cosmos/cosmos-sdk/blob/9fd866e3820b3510010ae172b682d71594cd8c14/proto/cosmos/staking/v1beta1/tx.proto#L120) and [MsgBeginRedelegate](https://github.com/cosmos/cosmos-sdk/blob/9fd866e3820b3510010ae172b682d71594cd8c14/proto/cosmos/staking/v1beta1/tx.proto#L104) in schema, and PR https://github.com/cosmos/cosmos-rust/pull/124, I remove `Option` from field amount for both messages.

### DIFF
--- a/cosmrs/src/staking.rs
+++ b/cosmrs/src/staking.rs
@@ -80,7 +80,7 @@ pub struct MsgUndelegate {
     pub validator_address: AccountId,
 
     /// Amount to UnDelegate
-    pub amount: Option<Coin>,
+    pub amount: Coin,
 }
 
 impl Msg for MsgUndelegate {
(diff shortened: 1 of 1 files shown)


tony-iqlusion (approved):
This looks good to me.

It's a breaking change though, so we'll need to merge it along with other breaking changes.

tony-iqlusion:
Thank you!

PR #197, opened 2022-04-05
Changed AccountId validation to check for maximum number of bytes

jstuczyn:
So that it would be compatible with the new 32-byte long addresses

### DIFF
--- a/cosmrs/src/base.rs
+++ b/cosmrs/src/base.rs
@@ -75,7 +75,7 @@ impl FromStr for AccountId {
     fn from_str(s: &str) -> Result<Self> {
         let (hrp, bytes) = bech32::decode(s).wrap_err("failed to decode bech32")?;
 
-        if bytes.len() <= MAX_ADDRESS_LENGTH {
+        if matches!(bytes.len(), 1..=MAX_ADDRESS_LENGTH) {
             Ok(Self {
                 bech32: s.to_owned(),
                 hrp_length: hrp.len(),
--- a/cosmrs/src/base.rs
+++ b/cosmrs/src/base.rs
@@ -6,6 +6,9 @@ use serde::{de, de::Error as _, ser, Deserialize, Serialize};
 use std::{fmt, str::FromStr};
 use subtle_encoding::bech32;
 
+/// Maximum allowed length (in bytes) for an address.
+pub const MAX_ADDRESS_LENGTH: usize = 255;
+
 /// Account identifiers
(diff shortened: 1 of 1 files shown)


cosmrs/src/base.rs
```diff
@@ -72,7 +72,7 @@ impl FromStr for AccountId {
     fn from_str(s: &str) -> Result<Self> {
         let (hrp, bytes) = bech32::decode(s).wrap_err("failed to decode bech32")?;
 
-        if bytes.len() == tendermint::account::LENGTH {
+        if bytes.len() >= tendermint::account::LENGTH {
```
tony-iqlusion, line 75: Seems like this should perhaps check inclusion within a set of allowed lengths?

```suggestion
        if matches!(bytes.len(), tendermint::account::LENGTH | 32) {
```

(seems like perhaps there should be a new upstream constant for 32-byte lengths, but we can cross that bridge when we get there)

jstuczyn: Well, I guess the decision is ultimately up to you. I was inspired by the associated error message, i.e. `account ID should be at least {} bytes long, but was {} bytes long` implying that the check should be performed against a range rather than against a set of allowed values. Let me know which one you prefer and I'll update the PR accordingly!

jstuczyn: Furthermore, I just looked at `cosmos-sdk` code directly and it looks like for their `AccAddress` (which I think the `AccountId` corresponds to?) they only length check they perform is whether the bytes decoded from the bech32 representation are shorter than 255 (and obviously whether they're not empty) (https://github.com/cosmos/cosmos-sdk/blob/f133317cfe8b4723ccb5e3687ffc9ab71b52ea75/types/address.go#L148-L150)

jstuczyn: I've updated the PR accordingly. Presumably `new` and `to_bytes` would also need to be changed, however, those would be incompatible breaking changes and thus I didn't want to touch them now : )

cosmrs/src/base.rs
```diff
@@ -72,16 +75,16 @@ impl FromStr for AccountId {
     fn from_str(s: &str) -> Result<Self> {
         let (hrp, bytes) = bech32::decode(s).wrap_err("failed to decode bech32")?;
 
-        if bytes.len() == tendermint::account::LENGTH {
+        if bytes.len() <= MAX_ADDRESS_LENGTH {
```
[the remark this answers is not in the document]
jstuczyn: I love it

PR #203, opened 2022-04-20
TryFrom tendermint public key types

cbrit:


cosmrs/src/crypto/public_key.rs
```diff
@@ -1,6 +1,7 @@
 //! Public keys
 
 use crate::{prost_ext::MessageExt, proto, AccountId, Error, ErrorReport, Result};
+use cosmos_sdk_proto::cosmos;
```
tony-iqlusion, line 4: `cosmos_sdk_proto` is imported as `proto` above, so you don't need an import here.

Instead you can use `proto::cosmos`

cosmrs/src/crypto/public_key.rs
```diff
@@ -120,6 +121,24 @@ impl TryFrom<&Any> for PublicKey {
     }
 }
 
+impl TryFrom<cosmos_sdk_proto::cosmos::crypto::ed25519::PubKey> for PublicKey {
+    type Error = ErrorReport;
+
+    fn try_from(public_key: cosmos_sdk_proto::cosmos::crypto::ed25519::PubKey) -> Result<PublicKey> {
```
tony-iqlusion, line 127: As mentioned above:

```suggestion
impl TryFrom<proto::cosmos::crypto::ed25519::PubKey> for PublicKey {
    type Error = ErrorReport;

    fn try_from(public_key: proto::cosmos::crypto::ed25519::PubKey) -> Result<PublicKey> {
```

cosmrs/src/crypto/public_key.rs
```diff
@@ -120,6 +121,24 @@ impl TryFrom<&Any> for PublicKey {
...
+        let tm_key = tendermint::public_key::PublicKey::from_raw_secp256k1(&public_key.key);
+        tm_key.map(Into::into).ok_or_else(|| Error::Crypto.into())
```
tony-iqlusion, line 129: I think this might be a bit easier to read if you get rid of the intermediate variable:

```suggestion
        tendermint::public_key::PublicKey::from_raw_secp256k1(&public_key.key)
            .map(Into::into)
            .ok_or_else(|| Error::Crypto.into())
```

Same goes for below.

tony-iqlusion:
Test failure looks unrelated.

It's an ongoing problem with the Docker-based integration tests being flaky, although it's a big annoying it's resurfaced as I thought it was solved.

tony-iqlusion:
@cbrit also you should be able to use these `TryFrom` impls in the `TryFrom<&Any>` impl since they provide the needed functionality.

You can replace the entire `try_from` method body with something like this now:

```rust
match any.type_url.as_str() {
    Self::ED25519_TYPE_URL => proto::cosmos::crypto::ed25519::PubKey::decode(&*any.value)?.try_into(),
    Self::SECP256K1_TYPE_URL => proto::cosmos::crypto::secp256k1::PubKey::decode(&*any.value)?.try_into(),
    other => Err(Error::Crypto).wrap_err_with(|| format!("invalid type URL for public key: {}", other)),
}
```

cosmrs/src/crypto/public_key.rs
cbrit: Done

cosmrs/src/crypto/public_key.rs
cbrit: Done

cosmrs/src/crypto/public_key.rs
cbrit: Done

cbrit:
> @cbrit also you should be able to use these `TryFrom` impls in the `TryFrom<&Any>` impl since they provide the needed functionality.
> 
> You can replace the entire `try_from` method body with something like this now:
> 
> ```rust
> match any.type_url.as_str() {
>     Self::ED25519_TYPE_URL => proto::cosmos::crypto::ed25519::PubKey::decode(&*any.value)?.try_into(),
>     Self::SECP256K1_TYPE_URL => proto::cosmos::crypto::secp256k1::PubKey::decode(&*any.value)?.try_into(),
>     other => Err(Error::Crypto).wrap_err_with(|| format!("invalid type URL for public key: {}", other)),
> }
> ```

Sweet. Just did this.

cosmrs/src/crypto/public_key.rs
```diff
@@ -101,22 +102,38 @@ impl TryFrom<&Any> for PublicKey {
...
-                tendermint::PublicKey::from_raw_ed25519(&proto.key)
+                proto::cosmos::crypto::ed25519::PubKey::decode(&*any.value)?.try_into()
             }
             Self::SECP256K1_TYPE_URL => {
-                let proto = proto::cosmos::crypto::secp256k1::PubKey::decode(&*any.value)?;
-                tendermint::PublicKey::from_raw_secp256k1(&proto.key)
+                proto::cosmos::crypto::secp256k1::PubKey::decode(&*any.value)?.try_into()
             }
             other => {
                 return Err(Error::Crypto)
```
tony-iqlusion, line 113: You can remove the `return` here (clippy will probably complain)

cosmrs/src/crypto/public_key.rs
```diff
@@ -4,6 +4,7 @@ use crate::{prost_ext::MessageExt, proto, AccountId, Error, ErrorReport, Result}
 use eyre::WrapErr;
 use prost::Message;
 use prost_types::Any;
+use proto::cosmos;
```
tony-iqlusion, line 7: Personally I'd recommend not importing this and referring to it as `proto::cosmos::*` instead.

Inside of CosmRS it can get quite confusing whether you're referring to a protobuf type or a domain type since they have the same names, so using `proto::*` keeps it clear a type is a protobuf type.

cosmrs/src/crypto/public_key.rs
tony-iqlusion: Also I think it'd be good to preserve this:

```rust
Err(Error::Crypto).wrap_err_with(|| format!("invalid type URL for public key: {}", other)),
```

It ensures the error message contains the unrecognized type URL

tony-iqlusion (approved):

tony-iqlusion:
Thank you!

PR #210, opened 2022-04-27
cosmrs: add sealed `ParseOptional` trait; fix `Fee` parsing

tony-iqlusion:
Adds an extension trait to `String`/`str` for handling a common Cosmos SDK pattern: treating a string field as optional if it's empty, or otherwise parsing it to a particular type.

This pattern existed in the `cosmwasm` and `fee` modules, with similar boilerplate to handle it. As it were, the code in the `Fee` module was buggy.

This redundant code can be replaced with this simple extension trait.


tony-iqlusion:
This was found via parsing a Juno protobuf transaction. It would be good to add a regression test before merging.

gavindoughtie (approved):

PR #215, opened 2022-05-02
Bump tendermint-rs crates to v0.23.7

tony-iqlusion:
This includes the following additional dependency upgrades:

- `prost` v0.10
- `tonic` v0.7

Also bumps `COSMOS_SDK_REV` from v0.45.2 => v0.45.4

Note: the changes to non-generated files are contained within ee56ca725e4a63d9aacb6869c3b51da67239ab44


proto-build/src/main.rs
```diff
@@ -347,7 +345,6 @@ fn compile_ibc_protos_and_services(out_dir: &Path) {
     tonic_build::configure()
         .build_client(true)
         .build_server(false)
-        .format(true)
```
tony-iqlusion, line 350: Note: this option was removed in `tonic` v0.7, which now uses the `prettyplease` crate for formatting:

https://github.com/hyperium/tonic/commit/d6c0fc112b2288a080fd0a727453b24d666e3d79

This also results in slightly different formatting from before (hence `TONIC_CLIENT_ATTRIBUTE` needing to change)

jkilpatr (approved):

PR #230, opened 2022-05-16
cosmos-sdk-protos: add `grpc-transport` feature

### LINKED ISSUE #226: Support `cosmos-sdk-proto` without `transport` feature of `tonic`
To call gRPC methods from browsers with `grpc-web` using webassembly, we need to disable `transport` feature of `tonic` which is not currently exposed in `cosmos-sdk-proto`.

tony-iqlusion:
Closes #226

Adds an on-by-default `grpc-transport` feature which enables the corresponding `transport` feature in `tonic`. The main use case is supporting gRPC in WASM environments.

To make this work with generated code, the generated protos are patched to feature-gate the `tonic::transport`-dependent code.

### DIFF
--- a/proto-build/src/main.rs
+++ b/proto-build/src/main.rs
@@ -47,22 +47,6 @@ const TMP_BUILD_DIR: &str = "/tmp/tmp-protobuf/";
 /// Protos belonging to these Protobuf packages will be excluded
 /// (i.e. because they are sourced from `tendermint-proto`)
 const EXCLUDED_PROTO_PACKAGES: &[&str] = &["gogoproto", "google", "tendermint"];
-/// Regex for locating instances of `tendermint-proto` in prost/tonic build output
-const TENDERMINT_PROTO_REGEX: &str = "(super::)+tendermint";
-/// Attribute preceeding a Tonic client definition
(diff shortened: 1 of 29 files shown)


proto-build/src/main.rs
```diff
@@ -434,8 +440,13 @@ fn copy_and_patch(src: impl AsRef<Path>, dest: impl AsRef<Path>) -> io::Result<(
...
+    let contents = contents.replace(TONIC_CLIENT_ATTRIBUTE, &GRPC_CLIENT_ATTRIBUTES.join("\n"));
 
-    fs::write(dest, patched_contents)
+    // Feature gate gRPC client impl blocks which rely on `tonic::transport`
+    let contents = Regex::new(TONIC_CLIENT_IMPL)
+        .unwrap()
+        .replace_all(&contents, TONIC_CLIENT_REPLACEMENT);
+
+    // Write output to destination
+    fs::write(dest, &*contents)
```
tony-iqlusion, line 451: It might be worth refactoring how these transformations are applied to the generated protos, making it into an folding iterator over `Regex` and replacement string tuples that applies successive transformations

tony-iqlusion:
Went ahead and did the cleanup I proposed regarding the regex substitutions on the output

PR #234, opened 2022-05-25
Additional protobuf-equivalent types

jstuczyn:
Hi!

I've added additional concrete types for things we already had defined in protobuf files, in particular `Response` types for various `cosmwasm` requests. But also for `MsgData` and `TxMsgData` from `abci`, for which I created a separate module in `cosmrs` crate. I was not entirely sure whether this was  the most appropriate place for it.

From the interesting bits, I've added the following bit to `MsgData` and `TxMsgData`:
```rust
impl MsgData {
    pub fn try_decode_as<M: Msg>(&self) -> Result<M> {
        M::Proto::decode(&*self.data)?.try_into()
    }
}

impl TryFrom<tendermint::abci::Data> for TxMsgData {
    type Error = ErrorReport;

    fn try_from(data: tendermint::abci::Data) -> Result<TxMsgData> {
        let proto = proto::cosmos::base::abci::v1beta1::TxMsgData::decode(data.value().as_ref())?;
        proto.try_into()
    }
}
```

those allow me more easily decode `data` fields from transaction results, for example as follows (which was my original motivation for creating this PR):
```rust
let tx_hash = "FFEE0EF6FE512F34536D2CAAEE2D854BC9465FA7ED4E1D8A63F15E24F26BD6C9".parse().unwrap();
let client = tendermint_rpc::HttpClient::new("http://localhost:26657").unwrap();
let tx = client.tx(tx_hash, false).await.unwrap();

let tx_msg_data = TxMsgData::try_from(tx.tx_result.data).unwrap();
let decoded_result: cosmwasm::MsgExecuteContractResponse = tx_msg_data.data[0].try_decode_as().unwrap();
// for completion sake that prints MsgExecuteContractResponse { data: [0, 0, 0, 0, 0, 0, 0, 1] } and the data is exactly what I expected to obtain
println!("{:?}", decoded_result);
```

Edit: I also added further modules: `auth` and `vesting` for handling different types of `Account`, such as `proto: proto::cosmos::auth::v1beta1::BaseAccount` or `proto::cosmos::vesting::v1beta1::BaseVestingAccount`

### DIFF
--- a/cosmrs/src/abci.rs
+++ b/cosmrs/src/abci.rs
@@ -83,7 +83,7 @@ impl From<TxMsgData> for proto::cosmos::base::abci::v1beta1::TxMsgData {
 }
 
 #[derive(Copy, Clone, Debug, Serialize, Deserialize, Default, Eq, PartialEq)]
-struct GasInfo {
+pub struct GasInfo {
     /// GasWanted is the maximum units of work we allow this tx to perform.
     pub gas_wanted: Gas,
 
--- a/cosmrs/src/abci.rs
+++ b/cosmrs/src/abci.rs
@@ -6,13 +6,21 @@ use prost::Message;
 use serde::{Deserialize, Serialize};
 use tendermint::abci::Gas;
 
+/// MsgData defines the data returned in a Result object during message execution.
 #[derive(Clone, Debug, Eq, PartialEq, PartialOrd, Ord)]
 pub struct MsgData {
+    /// Incoming message type that emitted this result data, for example `"/cosmos.bank.v1beta1.MsgSend"`.
     pub msg_type: String,
+
+    /// Binary data emitted by this message.
+    // Do note that usually the data has to be decoded into the corresponding protobuf `Response` type.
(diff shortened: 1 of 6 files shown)


tony-iqlusion:
This looks good at first glance. I'll try to do a more in-depth review soon.

cosmrs/src/abci.rs
```diff
@@ -0,0 +1,112 @@
...
+impl From<TxMsgData> for proto::cosmos::base::abci::v1beta1::TxMsgData {
+    fn from(tx_msg_data: TxMsgData) -> Self {
+        proto::cosmos::base::abci::v1beta1::TxMsgData {
+            data: tx_msg_data.data.into_iter().map(Into::into).collect(),
+        }
+    }
+}
+
+#[derive(Copy, Clone, Debug, Serialize, Deserialize, Default, Eq, PartialEq)]
+struct GasInfo {
```
tony-iqlusion, line 86: Did you mean for this to be `pub`?

```suggestion
pub struct GasInfo {
```

jstuczyn:
I'm not sure how I missed the `pub` on `GasInfo`. Regardless, I've fixed it and also added `rustdoc` comments on all structs (and fields) created

PR #237, opened 2022-06-01
Add bank::MsgMultiSend support

cbrit:


cosmrs/src/bank.rs
```diff
@@ -60,3 +60,142 @@ impl From<&MsgSend> for proto::cosmos::bank::v1beta1::MsgSend {
...
+        proto::cosmos::bank::v1beta1::MsgMultiSend {
+            inputs: msg.inputs.iter().map(Into::into).collect(),
+            outputs: msg.outputs.iter().map(Into::into).collect(),
+        }
+    }
+}
+
+/// Represents a MultiSend Input or Output
+#[derive(Clone, Debug, Eq, PartialEq, PartialOrd, Ord)]
+pub struct MultiSendIO {
```
tony-iqlusion, line 122: Per Rust naming conventions, I think this should be `MultiSendIo`:

```suggestion
pub struct MultiSendIo {
```

https://rust-lang.github.io/api-guidelines/naming.html

> In UpperCamelCase, acronyms and contractions of compound words count as one word: use Uuid rather than UUID, Usize rather than USize or Stdin rather than StdIn. In snake_case, acronyms and contractions are lower-cased: is_xid_start.

See also types like `IoSlice` in `std` for precedent.

tony-iqlusion:
@cbrit also if you rebase, I've added a `missing_docs` lint in #238

cbrit:
> @cbrit also if you rebase, I've added a `missing_docs` lint in #238

Done!

PR #250, opened 2022-06-29
Add feegrant module support

neacsu:
Expose MsgGrantAllowance, MsgRevokeAllowance, BasicAllowance and
AllowedMsgAllowance from cosmrs.

The current prost-types dependency version doesn't support proper
Duration error propagation, which should be fixed when
https://github.com/tokio-rs/prost/pull/625 is included in a release.
~~Until then, I only included the code for PeriodicAllowance as commented.~~
Created tracking issue for when a new prost-types release happens: https://github.com/cosmos/cosmos-rust/issues/252


cosmrs/src/feegrant.rs
```diff
@@ -0,0 +1,321 @@
...
+//     fn from(allowance: &PeriodicAllowance) -> proto::cosmos::feegrant::v1beta1::PeriodicAllowance {
+//         proto::cosmos::feegrant::v1beta1::PeriodicAllowance {
+//             basic: allowance.basic.map(Into::into),
+//             period: allowance.period.map(Into::into),
+//             period_spend_limit: allowance.period_spend_limit.iter().map(Into::into).collect(),
+//             period_can_spend: allowance.period_can_spend.iter().map(Into::into).collect(),
+//             period_reset: allowance.period_reset.map(Into::into),
+//         }
+//     }
+// }
```
tony-iqlusion, line 261: I would suggest removing this an opening a tracking issue, rather than checking in commented-out code

tony-iqlusion (changes requested):
One nit about commented-out code, otherwise this looks good

tony-iqlusion (approved):

PR #254, opened 2022-07-22
cosmrs: add `tx::BodyBuilder`

tony-iqlusion:
Adds a builder type for constructing `tx::Body`.

### DIFF
--- a/cosmrs/src/tx/builder.rs
+++ b/cosmrs/src/tx/builder.rs
@@ -1,0 +1,73 @@
+//! Transaction builder.
+
+use super::Body;
+use crate::Any;
+use tendermint::block;
+
+/// Transaction [`Body`] builder which simplifies incrementally assembling and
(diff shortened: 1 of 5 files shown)


cosmrs/src/tx/builder.rs
```diff
@@ -0,0 +1,70 @@
...
+
+    /// Set the transaction memo.
+    // TODO(tarcieri): error if the memo is already set?
+    pub fn memo(&mut self, memo: impl Into<String>) -> &mut Self {
+        self.body.memo = memo.into();
+        self
+    }
+
+    /// Set the timeout height.
+    // TODO(tarcieri): error if the timeout height is already set?
```
cbrit, line 41: fwiw, erroring if already set seems too opinionated to me

tony-iqlusion: Removed the TODOs

PR #264, opened 2022-07-27
Fix prefix address checks

bdeneux:
Based on the [bech32 specification](https://github.com/bitcoin/bips/blob/master/bip-0173.mediawiki#bech32), address prefix must contain 1 to 83 US-ASCII characters, with each character having a value in the range [33-126]. 

Following subtle-encoding [implementation](https://docs.rs/subtle-encoding/latest/src/subtle_encoding/bech32.rs.html#192-197) of bech32, we have fixed the prefix verification when create a new `AccountId` according to this spec.

### DIFF
--- a/cosmrs/src/base.rs
+++ b/cosmrs/src/base.rs
@@ -25,8 +25,7 @@ impl AccountId {
     pub fn new(prefix: &str, bytes: &[u8]) -> Result<Self> {
         let id = bech32::encode(prefix, &bytes);
 
-        // TODO(tarcieri): ensure this is the proper validation for an account prefix
-        if !prefix.chars().all(|c| matches!(c, 'a'..='z')) {
+        if !prefix.chars().all(|c| matches!(c, '!'..='@' | 'A'..='Z' | '['..='`' | 'a'..='z' | '{'..='~')) {
             return Err(Error::AccountId { id })
                 .wrap_err("expected prefix to be lowercase alphabetical characters only");
         }
--- a/cosmrs/src/base.rs
+++ b/cosmrs/src/base.rs
@@ -234,6 +234,13 @@ mod tests {
     }
 
     #[test]
+    fn account_id_with_digit() {
+        "okp41urdh3smlstyafjtyg0d606egllhwp8kvnw0d2f"
+            .parse::<AccountId>()
+            .unwrap();
+    }
+
+    #[test]
     fn denom_from_str() {
         assert!(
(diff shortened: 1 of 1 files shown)


tony-iqlusion:
This check is redundant with the checks already performed during Bech32 decoding. The purpose of the existing check is to enforce a subset of what Bech32 allows.

Is there a specific use case you have in mind for this?

bdeneux:
Yes, it's because when you try to create a new `AccountId` with a valid address prefix (following specification), there is some case that AccountId return Error of malformed prefix. This is our case because our address prefix contains a number while it is a valid address, it is impossible to create a new `AccountId`.

You're right, prefix verification is done at the [decoding part of Bech32](https://github.com/cosmos/cosmos-rust/blob/ab33607f6d899e961ee06ae7fb933e76c64b8c3e/cosmrs/src/base.rs#L85) when parsing from string, so, to avoid redundant check, remove this check from the new AccountId method could be a solution but if you create AccountId without parsing from string, error will be throw late (at decoding). 
This verification could also be the responsibility of the bech32 library, in the encoding part.

tony-iqlusion:
We could potentially relax it to allow alphanumeric prefixes. You're certain these work with the upstream Cosmos SDK? (i.e. they're allowed in a denom?)

bdeneux:
Yes 👍, we have a working cosmos blockchain using [`okp4` address prefix](https://github.com/okp4/okp4d/blob/main/app/app.go#L116). Address and transaction work successfully. https://explore.okp4.network/OKP4/account/okp41dfpnlvcaurtzlpxh3ek3gt303mu4q0hhhjflye

cosmrs/src/base.rs
```diff
@@ -25,10 +25,12 @@ impl AccountId {
     pub fn new(prefix: &str, bytes: &[u8]) -> Result<Self> {
         let id = bech32::encode(prefix, &bytes);
 
-        // TODO(tarcieri): ensure this is the proper validation for an account prefix
-        if !prefix.chars().all(|c| matches!(c, 'a'..='z')) {
+        if !prefix
+            .chars()
+            .all(|c| matches!(c, '!'..='@' | 'A'..='Z' | '['..='`' | 'a'..='z' | '{'..='~'))
```
tony-iqlusion, line 30:
```suggestion
            .all(|c| matches!(c, 'A'..='Z' | 'a'..='z' | '0'..='9'))
```

PR #271, opened 2022-08-08
Update tendermint-proto dependency to a version that works on wasmd

nicolaslara:
The current version of tendermint-proto used by cosmos-sdk-proto has a bad dependency that causes cosmwasm contracts that use cosmos-sdk-proto to not be allowed by wasmd. See https://github.com/informalsystems/tendermint-rs/blob/main/.changelog/v0.23.9/summary.md for the changelog and https://github.com/osmosis-labs/swaprouter/pull/1 for more details about the issue. 

Upgrading a minor version of the dependency fixes this.

### DIFF
--- a/cosmos-sdk-proto/Cargo.toml
+++ b/cosmos-sdk-proto/Cargo.toml
@@ -18,7 +18,7 @@ rust-version = "1.56"
 [dependencies]
 prost = "0.10"
 prost-types = "0.10"
-tendermint-proto = "=0.23.8"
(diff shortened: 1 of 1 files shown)


tony-iqlusion:
There's been a new release of all of the tendermint-rs crates, and they all need to be updated together (they use a weird versioning scheme where minor versions are a breaking change, which will hopefully go away soon).

I'll do that later today.

tony-iqlusion:
Closing in favor of #277

Issue #252, opened 2022-07-04
Track `prost-types` release for `Duration` error propagation


neacsu:
The current `prost-types` dependency version doesn't support proper Duration error propagation, which should be fixed when https://github.com/tokio-rs/prost/pull/625 is included in a release.

Once that is done, we can also expose `PeriodicAllowance`, similarly to how we did in https://github.com/cosmos/cosmos-rust/pull/250 for `MsgGrantAllowance`, `MsgRevokeAllowance`, `BasicAllowance` and
`AllowedMsgAllowance`

tony-iqlusion:
`prost` was updated to v0.11 in #277

PR #280, opened 2022-08-08
cosmrs: add additional `feegrant` domain types

tony-iqlusion:
Adds the domain types originally from #250 (removed in f63ded6) which were previously blocked on #252.

cc @neacsu

### DIFF
--- a/cosmrs/src/feegrant.rs
+++ b/cosmrs/src/feegrant.rs
@@ -2,10 +2,9 @@
 //!
 //! <https://docs.cosmos.network/master/modules/feegrant/>
 
-use std::time::SystemTime;
-
 use crate::{proto, tx::Msg, AccountId, Coin, ErrorReport, Result};
 use prost_types::Any;
+use std::time::{Duration, SystemTime};
 
 /// MsgGrantAllowance adds permission for Grantee to spend up to Allowance
 /// of fees from the account of Granter.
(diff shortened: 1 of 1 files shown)


cosmrs/src/feegrant.rs
```diff
@@ -178,6 +177,78 @@ impl From<&BasicAllowance> for proto::cosmos::feegrant::v1beta1::BasicAllowance
...
+
+impl From<PeriodicAllowance> for proto::cosmos::feegrant::v1beta1::PeriodicAllowance {
+    fn from(allowance: PeriodicAllowance) -> proto::cosmos::feegrant::v1beta1::PeriodicAllowance {
+        proto::cosmos::feegrant::v1beta1::PeriodicAllowance {
+            basic: allowance.basic.map(Into::into),
+            period: allowance
+                .period
+                .map(TryInto::try_into)
+                .transpose()
+                .expect("invalid allowance period"), // TODO(tarcieri): fallible serialization?
```
tony-iqlusion, line 240: @neacsu this is somewhat annoying... there's only `TryFrom` conversions between `prost_types::Duration` and `std::time::Duration`.

To get rid of the `expect` we'd need to switch the bound on `Msg::Proto` to be `TryInto` instead of `Into`. That's probably a good idea in general, but something I don't want to do as part of this release.

PR #289, opened 2022-08-31
Update cosmos-sdk-proto to v0.46 definitions

cbrit:
This is a manual process since third party proto deps have to be copied in to the cosmos-sdk-go git submodule. I also had to comment out submodule updates for the ibc-go and wasmd stuff becauset hey were resetting the cosmos-sdk-go submodule to an older commit, so their definitions are not updated.


tony-iqlusion:
Closing in favor of #395

PR #301, opened 2022-10-29
Allow bech32 String to be moved out of AccountId

de-husk:
All of the cosmos-sdk proto types use Strings for the account addresses: [ex](https://github.com/cosmos/cosmos-rust/blob/main/cosmos-sdk-proto/src/prost/cosmos-sdk/cosmos.bank.v1beta1.rs#L429). 

I know there is the `AsRef<str>` and `fmt::Display` impls, but I wanted a way to move the bech32 String out of the AccountId instead of having to allocate a new String each time.

Let me know if you think there's a better way.

### DIFF
--- a/cosmrs/src/base.rs
+++ b/cosmrs/src/base.rs
@@ -86,6 +86,12 @@ impl FromStr for AccountId {
     }
 }
 
+impl From<AccountId> for String {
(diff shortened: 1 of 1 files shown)


tony-iqlusion:
Thanks!

PR #306, opened 2022-11-29
Bump tendermint-rs crates to 0.27

tomtau:


tony-iqlusion:
@tomtau some errors when building `cosmrs` with the `rpc` feature.

If you'd like I can take care of those, although the solution should just be switching to `tendermint::Hash` everywhere.

tomtau:
@tony-iqlusion it should be fixed now

tony-iqlusion:
Thanks!

PR #314, opened 2023-01-02
Bump tendermint-rs crates to 0.28

tony-iqlusion:


tony-iqlusion:
@thanethomson this seems like a potential tendermint-rs bug?

https://github.com/cosmos/cosmos-rust/actions/runs/3825434673/jobs/6508425796#step:4:194

> error[E0277]: the trait bound `alloc::vec::Vec<u8>: core::convert::From<bytes::Bytes>` is not satisfied
>   --> /home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/tendermint-0.28.0/src/hash.rs:237:26
>    |
> 237 |         Ok(AppHash(value.into()))
>    |                          ^^^^ the trait `core::convert::From<bytes::Bytes>` is not implemented for `alloc::vec::Vec<u8>`

tony-iqlusion:
@thanethomson it looks like you're depending on a feature of `bytes` which was added some time after 1.1.

You might want to pin to `bytes` 1.3 (or whatever version added that `From` conversion)

thanethomson:
Thanks for bringing this up @tony-iqlusion! cc @mzabaluev

PR #360, opened 2023-03-19
Update type_urls.rs, added missing types.

Philipp-Sc:
added missing proposal types as well as others.


cosmos-sdk-proto/src/type_urls.rs
```diff
@@ -9,6 +9,83 @@ use crate::{cosmos, ibc, traits::TypeUrl};
 #[cfg(feature = "cosmwasm")]
 use crate::cosmwasm;
 
+#[cfg(feature = "cosmwasm")]
```
tony-iqlusion, line 12: I think it would make sense to logically group these with the other `cosmwasm`-gated impls.

It might even be worth considering a submodule

Philipp-Sc: yes, also good would be for the proto-build to auto generate the types, in the worst case via text parsing that should not be that hard, anyways I manually went over all the current governance proposal types and for now I have everything I need.

tony-iqlusion:
> yes, also good would be for the proto-build to auto generate the types

We could do that, although what I'd really love is for this to get fixed upstream in Prost:

https://github.com/tokio-rs/prost/issues/299

Issue #397, opened 2023-04-23
Enable cosmrs support `no_std`


DaviRain-Su:

DaviRain-Su:
I am implement ics06-solomachain-client, that use cosmrs publickey type. I wish cosmrs can implement `no-std`.

tony-iqlusion:
Adding `no_std` support is something I don't have time to work on, and ideally it wouldn't require a bunch of compromises on the error types like what happened with `tendermint-rs`.

DaviRain-Su:
Oh, there is no specific `no-std` platform support because I need to use types from `cosmrs` while implementing the `ICS06 solomachine client`. I don't think it makes sense to redundantly define which type conversions are needed in `ibc-rs`.

tony-iqlusion:
Note: `cosmos-sdk-proto` support for `no_std` was added in #478

tony-iqlusion:
Rust 1.81 is out and `core::error::Error` is now stable, which should unblock this issue

Issue #398, opened 2023-04-23
Maybe Can Use only one crate `cosmos-sdk-proto` or `ibc-proto-rs`


DaviRain-Su:
this is ibc-proto-rs: https://github.com/cosmos/ibc-proto-rs/tree/main

tony-iqlusion:
I have no idea what you're asking here

DaviRain-Su:
And cosmos-sdk-proto use different cosmos-sdk, ibc-go version with ibc-proto. and also problem is cosmos-sdk-proto is not support `no-std`.

tony-iqlusion:
It seems like you should file this issue against `ibc-proto-rs` instead of here? I'm not sure what we can do per se.

DaviRain-Su:
I am wondering if it is possible to merge the files generated by `cosmos-sdk-proto` into `ibc-proto-rs`.

tony-iqlusion:
Seems like you should open an issue at https://github.com/cosmos/ibc-proto-rs/issues then

PR #404, opened 2023-05-11
enable `no-std` and replate `cosmos-sdk-proto` to `ibc-proto`

DaviRain-Su:
close: #397 #398 

At present, one of the issues that needs to be addressed is to ensure that the versions of the proto files generated for `cosmos sdk`, `ibc`, and `wasmd` in `ibc-proto` correspond to the current library.

The overall changes have been made to support `no-std` in cosmrs, replaced `cosmos-sdk-proto` with `ibc-proto`. As for the error handling, I'm not sure if it meets your requirements or if you have any suggestions.


tony-iqlusion:
This PR seems to be combining a number of different changes, including rather huge, sweeping, and controversial ones like completely deleting `cosmos-sdk-proto`.

I think something like retiring `cosmos-sdk-proto` and naming `ibc-proto` as its official successor is a much, much bigger discussion, and especially shouldn't be mixed into other changes like trying to add `no_std` support to `cosmrs`.

PR #405, opened 2023-05-12
Add cosmos signing type

DaviRain-Su:


tony-iqlusion:
Sorry for not getting to this sooner! I removed the `TypeUrl` trait in #432 and so things are in a bit of a flux.

This will need to use the new `Name` trait although there's currently a bug that's being worked around making the boilerplate a bit nasty. Hopefully support for autogenerating `Name` trait impls with `prost-build` will happen soon.

PR #432, opened 2023-10-03
cosmos-sdk-proto: replace `TypeUrl` with `Name` trait

tony-iqlusion:
The `Name` trait added upstream in `prost` can be used to compute the type URL using `Name::type_url`: https://github.com/tokio-rs/prost/pull/896

The new `Any::{from_msg, to_msg}` methods can be used to serialize `Message` types to/from messages.

This commit switches from the `TypeUrl` trait to `Name` and relies more on upstream functionality in `prost`.

Unfortunately it shipped with a bug in the default `Name::full_name` method, so a workaround is temporarily used until a fix ships: https://github.com/tokio-rs/prost/pull/923


Kayanski:
Is this reversed now that prost-types has had a named release ?

PR #446, opened 2023-10-30
Add more name traits

penso:
I needed a few more name traits which might be useful for others


tony-iqlusion:
These look fine?

Also hopefully they can be autogenerated soon: https://github.com/tokio-rs/prost/pull/931

penso:
> These look fine?

Yes but I just wrote a macro for a better PR. I noticed they'd be automatically generated but I needed it now for my Cosmos indexer so I added them anyway.

Another PR is coming.

PR #447, opened 2023-10-31
Change name implementations to a macro

penso:
I needed a few more name traits and added a macro for easier management.


penso:
Not sure why the test with `cargo test --release --all-features` failed, works locally for me.

tony-iqlusion:
It's a spurious error in the integration test. I retried the build.

PR #457, opened 2024-01-16
Add more IBC related names

penso:
Adding a few missing type names implementations.

### DIFF
--- a/cosmos-sdk-proto/src/type_names.rs
+++ b/cosmos-sdk-proto/src/type_names.rs
@@ -28,11 +28,20 @@ impl_name!(
     "ibc.core.client.v1",
     "MsgUpdateClient"
 );
-
+impl_name!(
+    ibc::core::channel::v1::MsgChannelCloseConfirm,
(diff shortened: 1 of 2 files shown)


tony-iqlusion:
Really we should switch to generating this automatically, now that it's possible

penso:
> Really we should switch to generating this automatically, now that it's possible

Very much agree with this, as well as having serde added.

penso:
> Really we should switch to generating this automatically, now that it's possible

Can you (re)give me pointers to adding this automatically? I might look at it at some point if no one has done it.

tony-iqlusion:
@penso in `prost-build`, when we invoke `tonic_build::configure()`, it needs to be passed some additional configuration, which can be passed in by calling `.compile_with_config` rather than just `.compile`:

https://docs.rs/prost-build/latest/prost_build/struct.Config.html

The newly added `enable_type_names` option can do it:

https://docs.rs/prost-build/latest/prost_build/struct.Config.html#method.enable_type_names

Issue #458, opened 2024-01-19
Abandon `cosmos-sdk-proto` in favor of `ibc-proto`?


tony-iqlusion:
See issue here: https://github.com/cosmos/ibc-proto-rs/issues/182

I didn't create `cosmos-sdk-proto`, but where this repo was supposed to be a place for a multi-contributor collaborative effort, instead it's pretty much become a ghost town with me as the maintainer of last resort.

tony-iqlusion:
Closing in favor of #459

Issue #442, opened 2023-10-12
Update ibc-go types


srdtrk:
ibc-go types are outdated. For example, none of the packet data memo fields are included

tony-iqlusion:
Yeah, #423 is open with some updates.

I held off on doing those in the `cosmos-sdk-proto` v0.20 release because of all of the other changes and so we had at least one release with Tendermint v0.34 support that included a prost v0.12 upgrade.

That said, I can take a look at trying to rebase #423.

tony-iqlusion:
As of #492 we're now updated to the latest version: v8.4.0

Issue #83, opened 2021-05-30
adding support for json


leejw51crypto:
messages in the tx are encoded with protobuf,
need to know class first to decode by `typeurl`

if those messages can be parsed into serde_json, it would be easy to display
how about adding serde json support in compiling rust files from proto files?
(in prost)

tony-iqlusion:
We'd definitely like to add JSON support. Some complicating factors are a planned transition to "Proto JSON" from the previous "Amino JSON" was put on hiatus, or at least that's my understanding, but it may be out-of-date.

It'd be good to examine current and future strategies for JSON support in tendermint-rs so we can align with what they're doing, and ideally reuse some of their work:

https://github.com/informalsystems/tendermint-rs/issues/654

devashishdxt:
@tony-iqlusion We can use [`pbjson`](https://docs.rs/pbjson-build/latest/pbjson_build/) for this?

tony-iqlusion:
@devashishdxt unfortunately it seems like the JSON that's required (e.g. by `ibc-rs`/Hermes) differs from the standard Protobuf JSON, or at least that's my understanding.

See ample discussion on https://github.com/informalsystems/tendermint-rs/issues/654

scalalang2:
What if we add struct attributes ?
See this link : https://github.com/tokio-rs/prost/issues/75#issuecomment-365371756

tony-iqlusion:
If there's consensus that Cosmos SDK will eventually migrate to [Protobuf JSON](https://developers.google.com/protocol-buffers/docs/proto3#json), then I'm fine with using something like `pbjson` like @devashishdxt recommended.

hxrts:
relevant cosmwasm discussion https://github.com/CosmWasm/cosmwasm/issues/1448

ewoolsey:
Any updates on this? I'm trying to build a cosmos-ledger library and running into a wall. I need to be ale to convert an Any into JSON to display on the ledger, and can't figure out how to move forwards.

ewoolsey:
As a side note, is there currently any workaround I can use to get this functionality? I've been trying for an hour or two and I can't seem to come up with a way to get json from an Any, or any of the other message types for that matter.

tony-iqlusion:
`Any` contains a serialized proto, so you won't ever be able to just "convert it into JSON".

[...]

PR #459, opened 2024-01-30
Remove IBC protos in favour of the `ibc-proto` crate

romac:
Closes: https://github.com/cosmos/cosmos-rust/issues/458
Closes: https://github.com/cosmos/cosmos-rust/issues/442
Closes: https://github.com/cosmos/cosmos-rust/issues/83

As promised, here is my follow-up to https://github.com/cosmos/cosmos-rust/issues/458 and its [corresponding issue](https://github.com/cosmos/ibc-proto-rs/issues/182) in the ibc-proto-rs repository.

Here is the corresponding PR in `ibc-proto` which removes the generated Cosmos SDK protos and uses them via `cosmos-sdk-proto` instead: https://github.com/cosmos/ibc-proto-rs/pull/187

## Changes

- Remove ibc-go submodule and the corresponding generated proto definitions
- Update Cosmos SDK protos to v0.47.0
- Re-export ICS 23 proto definitions from the `ics23` crate instead of bundling them in `cosmos-sdk-proto`
- Add no-std compatible ProtoJSON (de)serialization to all Protobuf definitions, via [Informal's fork of `pbjson`](https://github.com/informalsystems/pbjson).

## TODO

- [x] Use `ics23` v0.11.1 once released (needed for `informalsystems-pbjson` v0.7.0)
  - [x] PR: https://github.com/cosmos/ics23/pull/274
  - [x] Release PR: https://github.com/cosmos/ics23/pull/275
- [x] Use `tendermint-proto` v0.34.1 once released (needed for missing serde annotations on some protos)
  - [x] PR: https://github.com/informalsystems/tendermint-rs/pull/1389
  - [x] Release PR: TBD

## Questions

- [ ] Do we want to update the Cosmos SDK protos further, eg. to v0.50.x?
- [ ] What's the best name for the feature guarding the ProtoJSON serialization code? Right now it's `serde` to match the corresponding feature in `ibc-proto`, but perhaps `pbjson` or `protojson` is a better name?

---

This PR is probably best reviewed commit by commit.

@tony-iqlusion This is by no mean the definitive way to go about deduplicating the work, so I am eager to hear what you think about this and to discuss further with you how best to proceeded.


cosmos-sdk-proto/Cargo.toml
```diff
@@ -16,18 +16,22 @@ edition = "2021"
 rust-version = "1.72"
 
 [dependencies]
+ics23 = "0.11.1"
 prost = "0.12"
-prost-types = "0.12"
+prost-wkt-types = "0.5"
```
tony-iqlusion, line 21: It seems like this crate requires `protoc`?

```
error: failed to run custom build command for `prost-wkt-types v0.5.0`

Caused by:
  process didn't exit successfully: `/home/runner/work/cosmos-rust/cosmos-rust/target/release/build/prost-wkt-types-d2782fb67e78a862/build-script-build` (exit status: 101)
  --- stderr
  thread 'main' panicked at 'Could not find `protoc` installation and this build crate cannot proceed without
      this knowledge. If `protoc` is installed and this crate had trouble finding
      it, you can set the `PROTOC` environment variable with the specific path to your
      installed `protoc` binary.If you're on debian, try `apt-get install protobuf-compiler` or download it from https://github.com/protocolbuffers/protobuf/releases
```

As things currently stand, both `tendermint-proto` and `cosmos-sdk-proto` pregenerate all the protos so downstream users don't need `protoc` installed in order to be able to use the crates.

So changing this would make `protoc` a hard dependency where it wasn't before.

romac: Ah sorry, didn't realize that as I have protoc installed locally. Let's then stick with `prost-types`.

romac: Ah my bad, I should have realized that and asked if it was okay. I actually don't see a good reason for moving to `prost-wkt-types` right now so I will revert that change so that we can keep pre-generating all the protos.

romac: Oh now I remember: Because some Protobuf messages contain `Any`, `Duration` or `Timestamp`, we need those so-called "well-known" types to derive `Serialize` in order to derive `Serialize` for the enclosing message via `pbjson`. Unfortunately, `prost-types` does not provide such an impl.

Potential solutions I see:
a) Use `prost-wkt-types` instead of `prost-types` and incur the cost of downstream users having to install `protoc`
b) Use `pbjson-types` instead of `prost-types`
c) Switch between `prost-types` and `pbjson-types` at compile time based on whether the `serde` feature is enabled or not

Would one of those be acceptable or do you see another solution?

romac: (d) Instruct `prost-build` to generate Rust definitions for the `google.protobuf` messages (ie. those "well-known types") instead of forwarding them to `prost-types` or another crate and let `pbjson-build` derive the `serde` impls for those.

tony-iqlusion: Do any of those have an integration with the `prost::Name` trait like `prost_types::Any::{from_msg, to_msg}` do? We use that a lot in `cosmrs`: https://docs.rs/prost-types/latest/prost_types/struct.Any.html#impl-Any

tony-iqlusion: I suppose if we generated our own Rust definitions for the `google.protobuf` messages, we could add `from_msg`/`to_msg` just like `prost-types` does.

tony-iqlusion: Perhaps to break this PR up into more manageable chunks, you could PR just this change? (and potentially other changes to `proto-build`)

romac: @tony-iqlusion Do we need any well-know types other than ~Duration`, `Timestamp` and `Any` for now?

tony-iqlusion: `Duration`, `Timestamp`, and `Any` sound fine to me

tony-iqlusion: Okay, we're now using `tendermint_proto::google::protobuf` types which support both `prost::Name` integrations and `informalsystems/pbjson`

tony-iqlusion:
@romac seems like you're unblocked on this if you'd like to try to move it forward

PR #460, opened 2024-02-21
cosmos-sdk-proto: update tonic to 0.11

alexheretic:
- Update `tonic` to v0.11

This is a breaking change to _cosmos-sdk-proto_.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -646,9 +646,9 @@ dependencies = [
 
 [[package]]
 name = "h2"
-version = "0.3.21"
+version = "0.3.24"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "91fc23aa11be92976ef4729127f1a74adf36d8436f7816b185d18df956790833"
+checksum = "bb2c4422095b67ee78da96fbb51a4cc413b3b25883c7717ff7ca1ab31022c9c9"
 dependencies = [
  "bytes",
  "fnv",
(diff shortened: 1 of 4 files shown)


alexheretic:
It would be helpful to publish a new _cosmos-sdk-proto_ release. Would you like me to update the versions in this PR?

Cargo.lock
```diff
@@ -1309,17 +1289,17 @@ dependencies = [
...
 name = "ring"
-version = "0.16.20"
+version = "0.17.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3053cf52e236a3ed746dfc745aa9cacf1b791d846bdaf412f60a8d7d6e17c8fc"
+checksum = "c17fa4cb658e3583423e915b9f3acc01cceaee1860e33d59ebae66adc3a2dc0d"
 dependencies = [
  "cc",
+ "cfg-if",
+ "getrandom",
```
tony-iqlusion, line 1298: Hmm, seems that *ring* is now pulling in `getrandom` which is breaking WASM builds.

alexheretic:
I can minimise the lockfile changes, if that helps. Though I don't understand what testing a lib with some old snapshot of historic dependencies achieves.

Cargo.lock
alexheretic: I removed the lockfile update, as it didn't directly relate to upgrading tonic.

PR #468, opened 2024-05-03
Update denom validation

Kayanski:
This PR aims at updating cosmrs to handle all valid coin denoms

## Context

With token factory modules, denoms are more diverse than before.
Denom validation must follow

Here is the Cosmos sdk coin validation function 
https://github.com/cosmos/cosmos-sdk/blob/94338e516d6fd095fdca175b47d87bd9020e7740/types/coin.go#L890


tony-iqlusion:
Good find on the upstream Cosmos SDK validation function... do you mind implementing the length check too, and removing the TODO and replacing it with a link to the upstream code? I can do it otherwise

Kayanski:
Hmm, letting you do this one if that's ok. Thanks

PR #471, opened 2024-06-01
cosmos-sdk-proto: add serde derive macros 

### LINKED ISSUE #409: Add additional derive macros to Prost generated types
If you take a look at [osmosis-std](https://crates.io/crates/osmosis-std) You'll see they have derived some additional traits on many of their types, which allow you get access to type urls, and request responses automatically. This is **_super_** powerful, because then you can automatically generate utility functions to send/receive messages using a client. This is what the generated code looks like:

```rust
/// QueryParamsRequest is the request type for the Query/Params RPC method.
#[allow(clippy::derive_partial_eq_without_eq)]
#[derive(
    Clone,
    PartialEq,
    Eq,
    ::prost::Message,
    ::serde::Serialize,
    ::serde::Deserialize,
    ::schemars::JsonSchema,
    CosmwasmExt,
)]
#[proto_message(type_url = "/cosmos.auth.v1beta1.QueryParamsRequest")]
#[proto_query(
    path = "/cosmos.auth.v1beta1.Query/Params",
    response_type = QueryParamsResponse
)]
pub struct QueryParamsRequest {}
```

The macros are already built so they could simply be used in this crate almost without changes.

ash-burnt:
Hello, 

I am currently trying to use cosmos protos for storing information inside of cosmwasm smart contracts, but the existing definitions don't have the required Serialize and Deserialize traits. This should fix #409 and #83 

I have extended the proto-build buf configuration to run an additional neoeinstein plugin to generate these traits.


cosmos-sdk-proto/Cargo.toml
```diff
@@ -19,6 +19,9 @@ rust-version = "1.72"
 prost = "0.12.3"
 prost-types = "0.12"
 tendermint-proto = "0.37"
+serde = "1.0.203"
+pbjson = "0.6.0"
+pbjson-types = "0.6.0"
```
tony-iqlusion, line 24: These don't support [`Any::from_msg`](https://docs.rs/prost-types/latest/prost_types/struct.Any.html#method.from_msg) and [`Any::to_msg`](https://docs.rs/prost-types/latest/prost_types/struct.Any.html#method.to_msg) which are used quite a bit in downstream code (at least mine) beyond this repo.

Per [this discussion](https://github.com/cosmos/cosmos-rust/pull/459/files/3d574b0fbced3f4dee9fdcee9dd338efacc99d4c#diff-2566f86e46e2addbeabb507d6df353b42a4017362e4626c7c501f77d0aa27439) the best way forward here might be to generate our own `google.protobuf` types and use those

tony-iqlusion: Here's an issue tracking that. We can hopefully get `Any::{from_msg, to_msg}` and `pbjson` both supported in tendermint-rs itself: https://github.com/informalsystems/tendermint-rs/issues/1445

tony-iqlusion:
@romac whatever is used to solve https://github.com/informalsystems/tendermint-rs/issues/1445 we can use here too

cosmos-sdk-proto/Cargo.toml
tony-iqlusion: #482 switches to using the new `tendermint-rs` definitions for these types which support the `no_std`-friendly fork of `pbjson`

tony-iqlusion:
@ash-burnt now that #482 is merged, if you rebase there should be a way forward here

cosmos-sdk-proto/Cargo.toml
```diff
@@ -17,7 +17,11 @@ rust-version = "1.72"
 
 [dependencies]
 prost = { version = "0.13", default-features = false }
+prost-types = "0.13"
 tendermint-proto = "0.39"
+serde = "1.0.203"
+pbjson = { git = "https://github.com/informalsystems/pbjson.git", package = "informalsystems-pbjson"  }
+pbjson-types = { git = "https://github.com/informalsystems/pbjson.git", package = "informalsystems-pbjson-types"  }
```
tony-iqlusion, line 24: Instead of these, use https://docs.rs/tendermint-proto/latest/tendermint_proto/google/protobuf/index.html

proto-build/src/main.rs
```diff
@@ -20,13 +20,13 @@ use walkdir::WalkDir;
...
 /// The Cosmos SDK commit or tag to be cloned and used to build the proto files
-const COSMOS_SDK_REV: &str = "v0.46.15";
+const COSMOS_SDK_REV: &str = "v0.47.10";
 
 /// The Cosmos ibc-go commit or tag to be cloned and used to build the proto files
 const IBC_REV: &str = "v3.0.0";
 
 /// The wasmd commit or tag to be cloned and used to build the proto files
-const WASMD_REV: &str = "v0.29.2";
+const WASMD_REV: &str = "v0.45.0";
```
tony-iqlusion, line 29: Can you bump these in a separate PR? There's already quite a bit going on in this one

cosmos-sdk-proto/src/prost/cosmos-sdk/cosmos.app.v1alpha1.rs
```diff
@@ -112,7 +112,7 @@ pub struct ModuleConfig {
     /// config is the config object for the module. Module config messages should
     /// define a ModuleDescriptor using the cosmos.app.v1alpha1.is_module extension.
     #[prost(message, optional, tag = "2")]
-    pub config: ::core::option::Option<::tendermint_proto::google::protobuf::Any>,
+    pub config: ::core::option::Option<::pbjson_types::Any>,
```
tony-iqlusion, line 115: These should stay as they are. Instead of using `pbjson-types`, you can leverage the newly added `pbjson` support in `tendermint-proto` (which solves every downstream crate having to figure out this problem)

cosmos-sdk-proto/Cargo.toml
ash-burnt: Done, the extern_paths have been updated to point to the new tendermint rs version, and these deps have been removed

proto-build/src/main.rs
ash-burnt: I have rolled back the version updates

cosmos-sdk-proto/src/prost/cosmos-sdk/cosmos.app.v1alpha1.rs
ash-burnt: After updating the extern_paths, these now generate as they did before

cosmos-sdk-proto/Cargo.toml
```diff
@@ -17,7 +17,9 @@ rust-version = "1.72"
 
 [dependencies]
 prost = { version = "0.13", default-features = false }
+prost-types = "0.13"
```
tony-iqlusion, line 20: This shouldn't be necessary anymore

proto-build/buf.sdk.gen.yaml
```diff
@@ -3,7 +3,12 @@ plugins:
   - plugin: buf.build/community/neoeinstein-prost:v0.3.1
     out: .
     opt:
+      - compile_well_known_types
+#      - extern_path=.google.protobuf=::pbjson_types
```
tony-iqlusion, line 7: Looks like this can be removed

tony-iqlusion: So I guess the problem was I added the config to `tonic` instead of `prost`?

cosmos-sdk-proto/Cargo.toml
tony-iqlusion, line 20:
```suggestion
```

proto-build/buf.sdk.gen.yaml
ash-burnt: I suppose so?

tony-iqlusion:
@ash-burnt looks like there are conflicts that need resolved

ash-burnt:
@tony-iqlusion they should be addressed now

ash-burnt:
Looks like these serde generations still require pbjson for serialization and deserialization of non-proto values. Seems like numbers and binary types

cosmos-sdk-proto/Cargo.toml
```diff
@@ -18,6 +18,7 @@ rust-version = "1.72"
 [dependencies]
 prost = { version = "0.13", default-features = false }
 tendermint-proto = "0.39.1"
+serde = "1.0.203"
```
tony-iqlusion, line 21: To preserve `no_std` support you'll need to either do this or make `serde` an optional dependency:

```suggestion
serde = { version = "1.0.203", default-features = false }
```

tony-iqlusion:
@ash-burnt yes, the (informalsystems-)`pbjson` crate is fine, it's just `pbjson-types` we're trying to avoid in lieu of using the definitions in `tendermint-proto`

cosmos-sdk-proto/Cargo.toml
ash-burnt: Committed this

cosmos-sdk-proto/Cargo.toml
```diff
@@ -18,6 +18,8 @@ rust-version = "1.72"
 [dependencies]
 prost = { version = "0.13", default-features = false }
 tendermint-proto = "0.39.1"
+serde = { version = "1.0.203", default-features = false }
```
tony-iqlusion, line 21: Probably need `serde`'s `alloc` feature:

```suggestion
serde = { version = "1.0.203", default-features = false, features = ["alloc"] }
```

proto-build/buf.sdk.gen.yaml
tony-iqlusion: I can fix it up in a followup

cosmos-sdk-proto/Cargo.toml
```diff
@@ -18,7 +18,7 @@ rust-version = "1.72"
 [dependencies]
 prost = { version = "0.13", default-features = false }
 tendermint-proto = "0.39.1"
-serde = { version = "1.0.203", default-features = false }
+serde = { version = "1.0.203", default-features = false, features = ["alloc"] }
```
tony-iqlusion, line 21: Hmm, it seems `protoc-gen-prost-serde` assumes traits like `ToString` are in the prelude. I guess to get this over the line, let's use the default `serde` features:

```suggestion
serde = "1.0.203"
```

tony-iqlusion:
Aah, so the problem isn't `serde` itself, but `no_std` in `tendermint-proto`

I think we're going to need to add a feature for this and have it enable the `std` feature of `tendermint-proto`.

Either that, or we're going to need to add a set of imports to the top of each of the generated `.serde.rs` files in lieu of the prelude.

ash-burnt:
What should the feature be called? `serde`?

tony-iqlusion:
Sure, that's fine

ash-burnt:
Went with "serialization" since it can't conflict with an import name

cosmos-sdk-proto/Cargo.toml
```diff
@@ -21,13 +21,16 @@ tendermint-proto = "0.39.1"
...
+serde = { version = "1.0.203", optional = true, default-features = false, features = ["alloc"] }
+pbjson = { package = "informalsystems-pbjson", optional = true, version = "0.7" }
 
 [features]
 default = ["grpc-transport"]
 std = ["prost/std", "tendermint-proto/std"]
 grpc = ["std", "tonic"]
 grpc-transport = ["grpc", "tonic/transport"]
 cosmwasm = []
+serialization = ["serde", "tendermint-proto/std", "pbjson"]
```
tony-iqlusion, line 33: There's a lot of serialization going on in this crate, both protobuf and now JSON.

How about:

```suggestion
serde = ["dep:serde", "tendermint-proto/std", "pbjson"]
```

ash-burnt: Done, and adjusted the code gen

tony-iqlusion:
@ash-burnt needs rustfmt but otherwise looks good

ash-burnt:
@tony-iqlusion formatted, let me know if there is anything else that needs doing

proto-build/src/main.rs
```diff
@@ -379,6 +379,17 @@ fn copy_and_patch(src: impl AsRef<Path>, dest: impl AsRef<Path>) -> io::Result<(
...
+        (
+            "impl serde::Serialize for",
+            "#[cfg(feature = \"serde\")]\n\
+            impl serde::Serialize for",
+        ),
+        (
+            "impl<'de> serde::Deserialize<'de> for",
+            "#[cfg(feature = \"serde\")]\n\
+            impl<'de> serde::Deserialize<'de> for",
+        ),
```
tony-iqlusion, line 392: These can probably be gated on the module level but I don't see any reason to block the PR over that

tony-iqlusion (approved):

PR #482, opened 2024-08-08
Bump `tendermint-rs` dependencies to v0.39

tony-iqlusion:
Also regenerates protos.

This notably switches from sourcing the `google.protobuf` well-known types from the `prost-types` crate and using `tendermint-proto`'s definitions instead.


Cargo.toml
```diff
@@ -5,3 +5,6 @@ members = [
     "cosmos-sdk-proto",
     "cosmrs"
 ]
+
+[patch.crates-io]
+tendermint-proto = { git = "https://github.com/informalsystems/tendermint-rs" }
```
tony-iqlusion, line 10: This required a few upstream changes to `tendermint-proto` which should be released doon

PR #502, opened 2024-09-09
cosmos-sdk-proto: remove `ics23` dependency

tony-iqlusion:
As discussed in cosmos/ibc-proto-rs#187, removes this dependency so `ics-proto` can import it directly and we don't need to worry about keeping the two synchronized.


romac (approved):

tony-iqlusion:
@romac with this merged, I think I can cut a final `cosmos-sdk-proto` v0.25.0 release

romac:
Sounds good to me!

PR #513, opened 2024-11-08
fix(cosmos-sdk-proto): `no_std` compatible `serde` feature

### LINKED ISSUE #512: `cosmos-sdk-proto`'s `serde` feature is `no_std` incompatible
We noticed `ibc-rs`'s `no-std-check` are failing because of `cosmos-sdk-proto`'s feature flag set.

cosmos/ibc-rs#1375

PS. This also makes `ibc-proto-rs`'s `serde` feature `no_std` incompatible: cosmos/ibc-proto-rs#254

rnbguy:
fix #512

### DIFF
--- a/cosmos-sdk-proto/Cargo.toml
+++ b/cosmos-sdk-proto/Cargo.toml
@@ -30,7 +30,7 @@ std = ["prost/std", "tendermint-proto/std"]
 grpc = ["std", "tonic"]
 grpc-transport = ["grpc", "tonic/transport"]
 cosmwasm = []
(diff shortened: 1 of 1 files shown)


tony-iqlusion:
These are technically breaking changes, but I don't imagine they'd actually have real-world impacts.

rnbguy:
Thanks so much for the quick response and release ! :rocket:

Since you talked about breaking changes, I reviewed my changes once again -- and realized that, probably I should have added `pbjson/std` to the `std` feature set. I will let you decide on it.

